Issue #116, opened 2014-08-29
Record timezone audio recordings and audio events were created.


cofiem:
Right now this information has to be inferred from the site location, which is difficult (and not required to be present). Changing this also affects #22 #24.

cofiem:
Added in f19d90ebb89ef7b334558b659a577820283b57af:
 - Ability to store timezones for site & users
 - UTC offset for audio recordings

UTC offset for audio recording is not yet stored by harvester.

Issue #23, opened 2013-12-04
Inconsistent Date and Time encodings in audio files


atruskie:
The rails app uses the following format strings to encode a date into an audio file name ([here](https://github.com/QutBioacoustics/baw-server/blob/master/app/models/audio_recording.rb#L66) ):

    :date => a.recorded_date.strftime("%Y%m%d"),
    :time => a.recorded_date.strftime("%H%M%S"),

The majority of the files we have are encoded with a different format string:

     "%y%m%d"
     "%H%M"

Suggested changes
- [x] temporarily check for both formats
- [x] save new files with new format
- [ ] at some point create a script to cleanup old format
- [ ] eventually remove redundant code
- [x] all file name encoding and decoding should be refactored into a single testable unit

cofiem:
See https://github.com/QutBioacoustics/baw-private/issues/81

Issue #22, opened 2013-12-04
Encode file Audio file names with UTC time zones


atruskie:
At the moment we encode the creation date of audio files in their filename like so (with the 10+ time zone): 

    original_audio/63/631ed7d5-ef52-4aa6-9062-ffdd41256930_100810-1200.mp3

There are 220K files named this way due to legacy reasons.

The associated record for this file stored in the  database under `recorded_date` is 

    2010-08-10 02:00:00 (UTC)

While the `config.timezone` rails setting is on there is no problem. However, if we ever want to change the timezone we will no longer be able to reference the bulk of the files.

Suggested tasks:
 - [x] Encode new files with UTC dates (disambiguate with `Z` suffix on time)
 - [ ] Cleanup old files (convert their datestamps to UTC)
 - [x] When checking for the presence of an audio file, check for both the `config.timezone` variant and the UTC variant

cofiem:
See https://github.com/QutBioacoustics/baw-private/issues/81

PR #121, opened 2014-09-06
[WIP] Added user timezone setting (defaults to Brisbane).

### LINKED ISSUE #24: Remove timezone configuration for brisbane
Assigning the default timezone to the site as Brisbane (10+) as in (e2435cbe76e0088601fa202cab5ab9abb6c7b75d)[https://github.com/QutBioacoustics/baw-server/blob/e2435cbe76e0088601fa202cab5ab9abb6c7b75d/config/application.rb#L38] is terrible practice.

It was done to meet time constraints and so that old legacy problems were not a problem.

Removing this assumption will affect #22 and #23

cofiem:
WARNING: finding original audio recordings may not work with user time zone other than Brisbane - test this!

Relates to #22 #23 
Closes #24 #116


cofiem:
We need to record the utc offset and the time zone name. This will enable us to display the time in the time zone the audio recording was made. We do not want to display the time in the user's timezone - that makes no sense.

We will need to retroactively add utc offsets and time zone names to all recordings. @jwim1 will help with this.

cofiem:
Need to record utc offset at time of recording (as this could change due to changes in time zone e.g. DST start/end).
Need time to be able to know the time zone recording was made (for display and provenance).

PR #211, opened 2015-06-14
Fix audio event tagging

atruskie:
It was missing the most crucial field - the one that made adding in the
tagging relation all worth it!


atruskie:
Please deploy staging if this is good :smile:

cofiem:
My bad - I must have been confused by the `:id` field and thought that was the tag id.

PR #225, opened 2015-07-09
[WIP] Implementing server-side custom jobs 

cofiem:
See [Custom Jobs](https://github.com/QutBioacoustics/baw-server/milestones/Custom%20Jobs) milestone.


atruskie:
I've had a look through this. It is massive... no obvious defects that I can see. More than likely I'll only spot bugs as I user API

PR #284, opened 2016-04-13
Added analysis_job_id and audio_recording_id to results API

### LINKED ISSUE #282: Add fields to analysis results API
The analysis results API currently returns payloads like this:

`GET https://staging.ecosounds.org/analysis_jobs/system/audio_recordings/187684/`

```json
{
    "meta": {
        "status": 200,
        "message": "OK"
    },
    "data": {
        "path": "/",
        "name": "/",
        "type": "directory",
        "children": [
            {
                "name": "log.txt",
                "type": "file",
                "size_bytes": 196054,
                "mime": "text/plain"
            },
            {
                "name": "Towsey.Acoustic.yml",
                "type": "file",
                "size_bytes": 1968,
                "mime": "application/x-yaml"
            },
            {
                "path": "/Towsey.Acoustic",
                "name": "Towsey.Acoustic",
                "type": "directory",
                "has_children": true
            }
        ]
    }
}
```

Problem: the information provided does not provide enough information to recreate the URL to child resources. The information is in the URL but that can make it a little hard to access. Our other standard API objects generally include parent relationship foreign key ids.

Proposed solution: include `analysis_job_id` and `audio_recording_id` in the payload (top level object only).

atruskie:
Also needed to add support for modifying the options that were passed
into the rspec_api_documentation helper method
`standard_request_options`

Closes #282


atruskie:
Tests pass locally. Travis error seems to be because of https://github.com/brigade/haml-lint/blob/master/CHANGELOG.md#0130

spec/helpers/acceptance_spec_helper.rb
```diff
@@ -26,14 +26,21 @@ def document_media_requests
 # @option opts [Hash]    :file_exists            (nil) Check if file exists
 # @option opts [Class]   :expected_error_class   (nil) The expected error class
 # @option opts [Regexp]  :expected_error_regexp  (nil) The expected error regular expression
+# @param [Proc] opts_mod an optional block that is called when rspec is running - allows dynamic changing of opts with
+#    access to rspec context (i.e. let and let! values)
 # @return [void]
-def standard_request_options(http_method, description, expected_status, opts = {})
+def standard_request_options(http_method, description, expected_status, opts = {}, &opts_mod)
```
cofiem: I'm not so sure about this proc - a block might be better. I'll have a look at it.

atruskie: the proc is needed because AFAIK you can't pass a block through to another closure without an explicit name. Otherwise the implicit block of the closure would be referenced. `example` on line 37, has its own block. Using `yield` on 41 I think causes issues.

PR #285, opened 2016-04-13
chore: Updated development environment

atruskie:
- changed default paths in default.yml to isolate to environment
(otherwise tests wipe out caches for development environments)
- added a google maps key stub to prevent the script templater from
erroring
- changed seeds.rb so that whenever db:setup is run it force resets the
details (for those users in particular we want them to always have a
known state)
- Updated readme.md to deal with the realities of guard and filesystem
polling events
- Added style guide link to documentation
- changed the windows nfs config to map the nfs mount to the vagrant
user
- changed the vagrant VM memory default to 1.5GB of RAM (ran out of
space while running dev server and running tests)


db/seeds.rb
```diff
@@ -4,21 +4,30 @@
...
-  harvester = User.new(user_name: 'Admin',
-                       email: Settings.admin_user.email,
-                       password: Settings.admin_user.password)
-  harvester.roles = [:admin]
-  harvester.skip_confirmation!
-  harvester.save!(validate: false)
+if User.where(user_name: 'Admin').first.blank?
+  admin = User.new
+else
+  admin = User.where(user_name: 'Admin').first
```
cofiem: Why do the user attributes need to be reset every time the database is seeded?

atruskie: I changed this mainly to help me setup a database which I had copied off staging. My logic follows

- seed is only run at initial db:setup at the moment
- the password seed may not be the same, meaning logins won't work. Useful to reset these accounts
- changing the behavior makes db:seed useful for setting the state of the database to a desired state. Admin/harvester should always have the email and password from settings

I thought these changes were useful to keep. Main reason for this whole PR :smile:

Issue #286, opened 2016-04-19
Allow query string filter parameters to be used on field_mapping attributes


atruskie:
In the [API spec](https://github.com/QutBioacoustics/baw-server/wiki/Rails-API-Spec---Filtering), query string parameters (QSPs) are allowed to be used for filtering using equality only semantics.

It does not appear this functionality has been implemented for field mappings (which seem to me to be virtual fields). Other attribute types seem to work fine.

Example cases:

| Compare Type | Request | Works? |
|--------------------|-------------|----------|
| String  | `GET /scripts?filter_name="simulate work"` | :+1: | 
| Number | `GET /scripts?filter_version=2` | :+1: | 
| Field Mapping | `GET /scripts?filter_is_last_version=true` | :-1:  | 
| Number | `GET /audio_recordings/234234/audio_events?filter_start_time_seconds=0.1161` | :+1: | 
| Field Mapping | `GET /audio_recordings/234234/audio_events?filter_duration_seconds=1.8112` | :-1:  | 

_`is_last_version` is a field mapping from [here](https://github.com/QutBioacoustics/baw-server/commit/4db980b4531b65d47e730247b0e0c6b7067ec239#diff-9cacd63540753d6929377d6cb58210cbR107)_

The failed requests currently fail with a 500 because the field is passed through to the database (it shouldn't be allowed to do that either) and this error is raised: `ActiveRecord::StatementInvalid (PG::UndefinedColumn: ERROR:  column audio_events.duration_seconds does not exist`

Example source object for reference:
```
{
"id": 2,
"name": "simulate work VERSION 2",
"description": "simulates running an analysis",
"analysis_identifier": "SIMULATE_WORK",
"version": 2,
"group_id": 1,
"creator_id": 1,
"created_at": "2016-02-18T05:58:05.465Z",
"executable_settings": "# nothing\r\nsomeSettings:\r\n key: value\r\n aList:\r\n - a\r\n - b\r\n - c",
"executable_settings_media_type": "text/plain",
"is_last_version": true,
"is_first_version": false
}
```

Please add support for filtering by QSP for field mappings.

cofiem:
I've almost got this working. Then #287 should work

atruskie:
Cool, lots of lines removed. Nice work. :+1:

cofiem:
Ha, yup. Much of the added stuff is tests, too.

PR #287, opened 2016-04-19
Fixes #259 - Implements  `is_first_version` and `is_last_version` attributes

### LINKED ISSUE #259: Scripts API enhancements
Please expose the following in the scripts API:

- The #show endpoint
- The `executable_settings` field and (if #258 is a go) the `executable_settings_filename` field
- and _discuss_: should `/filter` return all versions of a script? It currently does.
  - personal preference is to not allow this... but the corollary is how do we allow seeing older versions?

atruskie:
Adds virtual fields `is_first_version` and `is_last_version` to the
Scripts API. Does NOT implement the `/groups` API endpoint - unnecessary
at this time.

Note one unit test currently fails because #286 is not implemented.


app/models/script.rb
```diff
@@ -27,51 +27,103 @@ def latest_version
...
+    joins(query).select('"scripts".*, "s2"."last_version", "s2"."first_version"')
+  end
+
+  # a fake reference to the aliased table defined in the default scope
+  @script_group = Arel::Table.new('s2', @arel_engine)
+
+  # HACK: reload item - this is a poor version of Persistence::reload
+  # https://github.com/rails/rails/blob/1f98eb60e59f4f70ef66ac2454ad029f46e3b27c/activerecord/lib/active_record/persistence.rb#L432
+  # The only difference is the `unscoped` method is removed.
+  def reload(options = nil)
```
cofiem: why is this needed?

atruskie: See:
- https://github.com/QutBioacoustics/baw-server/pull/287/files/4db980b4531b65d47e730247b0e0c6b7067ec239#diff-9cacd63540753d6929377d6cb58210cbR157
- https://github.com/QutBioacoustics/baw-server/pull/287/files/4db980b4531b65d47e730247b0e0c6b7067ec239#diff-42b71321a35f19e8409b3b39966e6f24R62
- https://github.com/QutBioacoustics/baw-server/pull/287/files/4db980b4531b65d47e730247b0e0c6b7067ec239#diff-42b71321a35f19e8409b3b39966e6f24R69

spec/acceptance/scripts_spec.rb
```diff
@@ -119,11 +210,38 @@ def scripts_id_param
...
         ],
-        data_item_count: 1,
+        data_item_count: 4,
         response_body_content: [
             '"analysis_identifier":{"contains":" identifier "}',
             '"executable_settings":"']
     })
   end
 
+  post '/scripts/filter' do
```
cofiem: Is this the test that fails due to #286?

atruskie: Nope, see https://github.com/QutBioacoustics/baw-server/pull/287/files/4db980b4531b65d47e730247b0e0c6b7067ec239#diff-944a313f082f8ee3273a4aea8688a805R73

atruskie:
So I'm reviewing the changes in 0ab8f91.

Converting the `last_version` and `first_version` properties to queries only is pretty bad for performance. A `GET /scripts` goes from 1 database query to **2×N + 1** queries - and that's ignoring the paging query! Essentially, the values in the default scope never get used.

Please consider the changes in 8bc7cb2 which use the attribute values if they are present and query for them in the rare case they are not present.

cofiem:
That looks reasonable - have you tested your most recent change much?

atruskie:
Unit and integration tests pass. Basic tests with http endpoint tester behave as I expect for `index`, `show`, and `filter` (with a `is_last_version` filter). `index` also works with `?filter_is_last_version=true` usecase.

cofiem:
cool, I'll review and merge tonight

Issue #192, opened 2015-05-07
Ensure it is possible to query/filter resque jobs based on some kind of field


atruskie:
Related to #191 and #193

For our chosen architecture, where we do not track sub-jobs, we need to be able to query resque to find resque jobs that have a custom_job's (i.e. the parent) id.

Whether this id is stored in the payload or in some other way remains unclear.

Implement such a filtering function

cofiem:
This issue is mainly about being able to report the number of resque queued, working, successful, and failed jobs for a particular `analysis_job`. One resque job processes a single `audio_recording`.

The way the counts are updated is complicated and prone to race conditions.
- (push) if workers update the stats, race conditions are very likely to occur. However, each worker just needs to send the analysis_id, and the status to increment. This remains ambiguous - if a failed status is sent, is working or queued decremented? the status per job is not stored.
- (pull) if stats are updated on demand or on a regular schedule by the web server, then the challenges below apply

This is possible in theory:
- queued jobs: filter the appropriate queue to find jobs matching an analysis_job id
- working jobs: get all the working jobs, check the analysis_job id
- failed jobs: check the failed queue for any jobs of the `Analysis` type with a matching analysis_job id
- successful jobs: total jobs pushed to the queue when the analysis_job was started - (queued jobs + working jobs + failed jobs)

In practice there are some challenges:
- the queued, working, and failed jobs lists have to be iterated, even if an operation that filters is used (e.g. [SCAN](http://redis.io/commands/scan). This could take a long time if there are lots of jobs. This operation could probably not be done regularly, or on demand.
- Since successful jobs are simply removed, there is no 'source of truth`: all values would be approximations.
- the same problems apply to finding a particular job. It is not possible to store information about individual jobs.

The other option is to have a job item table to store very basic information. This has the benefit of removing all the challenges of using redis only. However, it is duplicate information, and needs to be kept in sync with the resque jobs. This would probably be done through workers pushing information to the web server.

atruskie:
The way i see it there are three options (I'm rehashing some of what you said, just thinking, no decisions):

# options

## (push): 
  - description: workers update stats on `analysis_job` record through HTTP+REST
  - pros
     - simplicity
     - does not need a meta-tracker (see below)
  - cons
    - no tracking of job items
    - race conditions
    - not normal form
    - cannot restart a job item
  - opportunities
    - We could defensively code for race conditions. Something like (SQL pseudo code):

        ```
        START TRANSACTION;
[...]

PR #290, opened 2016-06-02
Feature: analysis jobs items crud

### LINKED ISSUE #274: Need to expose analysis results' audio recordings
Analysis jobs produce  results. For example `analysis_job.id == 1` may have results for recordings
`123,456,9439493,213131,122333,...`. It is quite reasonable that there could be 50K folders in the `.../analysis_results/1/` folder (one of our projects actually has that many recordings).

To expose results to users, so they can look/download/load up into tools, they need to be able to choose a recording to show the results. Recording IDs that are in the job are not guaranteed to follow any pattern (they are not sequential, often not contiguous). 

**Currently, the information for which audio recordings are part of a job is not tracked anywhere. These IDs are also not accessible via any part of the API**. Whatever solution chosen needs to support paging the results (i.e. some form of offset and limit).The following possible solutions exist:

1. native dir enumeration (i.e. `Dir.entries`)
  - pros: simplicity, extend virtual file system concept already used for analysis results
  - cons: directory listing for even a few hundred files is very slow. This effect is made worse because we rely on `noac` (no client cache) to ensure our nfs shares remain performant. Most importantly, I can't find a ruby API to page results.
1. tracking table
  - we'd use the job items concept from [here](https://github.com/QutBioacoustics/baw-server/issues/192#issuecomment-189828775) - specifically the (push+tracking table) idea
  - pros: next simplest idea, reuse a concept we're planning to make. Supports paging.
  - cons: this table could end up being massive (n^2 perf profile). We'd need to store the records permanently.
1. re-executing the saved saved search query
   - pros: possibly simple. we don't have to store anything. also supports paging.
   - cons: the results of the search could change when reexecuting. inconsistent results. On very large result sets the query time could prove to be prohibitive. Also the only item here that would return results for analysis results folders that don't exist yet.
1. caching saved search audio recording ids in simple binary manifest in `.../analysis_results/1/manifest.bin`
  - pros: divorced from database, storage of data isolated to analysis_jobs folder. essentially a cached `ls`
  - cons: would need to read and parse a file (very quickly) on every request - file could easily reach several hundred KB in size.
1. use swift (object store)
  - pros: it is made for this use case - hundreds of thousands of small files. It also supports paging.
  - cons: would introduce another dependency in our system. We'd likely need a fallback for disk based access anyway.

I'm currently unsure what the best path forward is.

Quick and dirty: option 1. and flag this as a potential future issue. :tada: technical debt.

Ideal: support 5. with a fallback for disc access based on 2.

Probably need to go with 2. Should be mediumly scalable.

atruskie:
This pull request supersedes #288

This feature branch adds in support for Analysis Jobs' Items. In particular, this branch has the code, logic, and tests, for CRUD operations. It DOES NOT include actual integration with Analysis Jobs; that is when an `AnalysisJob` is created it should also create `AnalysisJobsItem`s - that functionality will come in a later request.

This pull request also unifies the analysis results controller with the new `analysis_jobs_items` controller. All results, whether `system` or `AnalysisJob` generated are accessible via the same REST endpoint with same properties available.

This work lays out the foundation for resolving #192.
Fixes #274.


app/models/analysis_jobs_item.rb
```diff
@@ -0,0 +1,197 @@
...
+
+    super(new_status)
+
+    update_status_timestamps(self.status, old_status)
+
+  end
+
+  def analysis_job_id
+    super_id = super()
+    return SYSTEM_JOB_ID if !new_record? && id.nil?
```
atruskie: TODO add comment: this patches the analysis_job_id for fake AnalaysisjobsItem s returned by the fake query....

app/models/analysis_jobs_item.rb
```diff
@@ -0,0 +1,197 @@
...
+        valid_fields: fields,
+        render_fields: fields,
+        text_fields: [:queue_id],
+        controller: :analysis_jobs_items,
+        action: :filter,
+        defaults: {
+            order_by: :audio_recording_id,
+            direction: :asc
+        },
+        valid_associations: [
```
atruskie: TODO: consider adding more nested relations
NB: available set to false for many-many tables

app/models/analysis_jobs_item.rb
```diff
@@ -0,0 +1,197 @@
...
+  end
+
+  # Update status and modified timestamp if changes are made. Does not persist changes.
+  # @param [Symbol] new_status
+  # @param [Symbol] old_status
+  # @return [void]
+  def update_status_timestamps(new_status, old_status)
+    return if new_status == old_status
+
+    case
```
atruskie: replace all `and`s with `&&`

app/models/analysis_jobs_item.rb
```diff
@@ -0,0 +1,197 @@
...
+  # Scoped query for getting fake analysis_jobs.
+  # @return [ActiveRecord::Relation]
+  def self.system_query
+    analysis_jobs_items = self.arel_table
+
+    # alias audio_recordings so other add on queries don't get confused
+    audio_recordings_inner = Arel::Table.new(:audio_recordings).alias('tmp_audio_recordings_generator')
+
+    # get a list of all other columns - this ensures attributes don't raise a MissingAttributeException
+    columns = (AnalysisJobsItem.column_names - ['audio_recording_id']).map { |c| '"' + table_name + '"."' + c + '"' }
```
atruskie: use `analysis_jobs_items.table_name`

app/models/ability.rb
```diff
@@ -236,14 +241,14 @@ def to_bookmark(user)
...
-    # must have read permission or higher on all saved_search.projects to create analysis job
+    # must have read permission or higher on ANY saved_search.projects to create analysis job
     can [:show, :create], AnalysisJob do |analysis_job|
       check_model(analysis_job)
       fail CustomErrors::BadRequestError.new('Analysis Job must have a saved search.') if analysis_job.saved_search.nil?
       projects = analysis_job.saved_search.projects
       fail CustomErrors::BadRequestError.new('Saved search must have at least one project.') if projects.size < 1
 
-      Access::Check.can_all?(user, :reader, projects)
+      Access::Check.can_any?(user, :reader, projects)
```
atruskie: Add comment explaining permissions relations a->b->c etc...

Issue #305, opened 2016-11-07
AudioEvents filter failing with malformed SQL


atruskie:
```
An ActiveRecord::StatementInvalid occurred in audio_events#filter:

PG::InvalidTextRepresentation: ERROR:  invalid input syntax for type numeric: "---
:from: 0 :to: " LINE 1: …time_seconds" – “audio_events"."start_time_seconds”) >= '---

^
: SELECT COUNT(*) FROM “audio_events” INNER JOIN “audio_recordings” ON “audio_recordings"."id” = “audio_events"."audio_recording_id” AND ("audio_recordings"."deleted_at" IS NULL) INNER JOIN “sites” ON “sites"."id” = “audio_recordings"."site_id” AND ("sites"."deleted_at" IS NULL) WHERE ("audio_events"."deleted_at" IS NULL) AND (EXISTS (SELECT 1 FROM “projects_sites” WHERE “sites"."id” = “projects_sites"."site_id” AND EXISTS (( SELECT 1 FROM “projects” WHERE “projects"."deleted_at” IS NULL AND “projects"."creator_id” = 234 AND “projects_sites"."project_id” = “projects"."id” UNION ALL SELECT 1 FROM “permissions” WHERE “permissions"."user_id” = 234 AND “permissions"."level” IN ('reader', ‘writer’, ‘owner’) AND “projects_sites"."project_id” = “permissions"."project_id” ))) OR EXISTS (SELECT 1 FROM “audio_events” “ae_ref” WHERE “ae_ref"."deleted_at” IS NULL AND “ae_ref"."is_reference” = ‘t’ AND “ae_ref"."id” = “audio_events"."id")) AND “audio_events"."is_reference” = ‘t’ AND (("audio_events"."end_time_seconds” – “audio_events"."start_time_seconds”) >= '--- :from: 0 :to: ') AND ("audio_events"."low_frequency_hertz" >= 1100)

lib/modules/api/response.rb:272:in `add_paging_and_sorting'
```

The request came from the reference library: 
```
------------------------------- Request:

* URL        : https://www.ecosounds.org/audio_events/filter
* HTTP Method: POST
[...]

PR #309, opened 2016-12-03
validate filter values

cofiem:
See #305


atruskie:
@cofiem this looks great - thanks! 

Can you verify the associated documentation is still accurate? https://github.com/QutBioacoustics/baw-server/wiki/API:-Filtering
As far as I can tell it still is.

cofiem:
No worries. I added one more commit to make the class checks consistent.

Yes, the documentation is still correct. I updated it to add a line about the types of values expected.

Issue #302, opened 2016-08-15
Deleted audio events are being returned in download annotations link


atruskie:
Reported by @KarlinaInd, annotations downloaded in the CSV format include those that have been deleted. THIS SHOULD NOT HAPPEN.

Confirmed by downloading a fresh CSV and also querying the database. May be related to #292 

Update: this is likely caused by the `csv_query` method using the AREL table over the active record query with the default non-deleted scope. https://github.com/QutBioacoustics/baw-server/blob/develop/app/models/audio_event.rb#L119

cofiem:
:dash: yup, that's likely the issue

JessCappadonna:
Remember soon after I started I showed you guys links in the excel spreadsheet that no longer had annotations as well.... it was never resolved so I just delete them whenever I download the CSV

PR #310, opened 2016-12-04
exclude deleted annotations from csv download

cofiem:
see #302

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -96,6 +96,10 @@ gem 'jc-validates_timeliness', '~> 3.1.1'
 
 gem 'enumerize', '~> 2.0'
 gem 'uuidtools', '~> 2.1.5'
+
+# Note: if other modifications are made to the default_scope
+# there are manually constructed queries that need to be updated to match
+# (search for ':deleted_at' to find the relevant places)
 gem 'acts_as_paranoid', git: 'https://github.com/ActsAsParanoid/acts_as_paranoid.git', branch: :master, ref: 'c2db19554ddaedcac0a2b8d6a0563dea83c972c5'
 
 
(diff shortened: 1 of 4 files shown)


atruskie:
Looks good.

My only question: Does doing the deleted check like I did in https://github.com/QutBioacoustics/baw-server/blob/b6b48089a1f790a3d2874fe4d8137dcc0d4f0d0b/app/models/analysis_jobs_item.rb#L159 offer any real advantages?

cofiem:
Well, firstly, the comment "cast back to active record relation" is misleading - it isn't a cast, it's using the ActiveRecord [`from`](http://api.rubyonrails.org/classes/ActiveRecord/QueryMethods.html#method-i-from) with the Arel query as the query.

I think you need line 160. Without line 160 (`query.joins(:audio_recording)`), you wouldn't get the ActiveRecord scope, you'd only have the `from` subquery, so line 160 is necessary, yes.

atruskie:
I meant, for using in this fix... All the default scopes should be used, if they were this bug (#302) wouldn't have happened...

cofiem:
Oh, righto. I'm not sure. The Arel query might be too complex to be able to make use of ActiveRecord for parts of it. It's a good thought, tho.

JessCappadonna:
Hi guys, 
I am soooo excited this is happening. One thing I am wondering, can we still see deleted things if we want too? I was envisioning having a way to filter them out/sort, rather then deleting them all together. This isn't critical, but would be helpful to know if one person is changing what another has done.

atruskie:
@JessCappadonna this changeset will remain scoped to fixing the bug only (many small changes increase the speed at which we can ship fixes),  however, I've taken your comment and integrated it into ideas we have discussed previously. You can comment and review here: https://github.com/QutBioacoustics/baw-server/issues/311

atruskie:
This is ready to merge.

I have an idea to improve it. If I find the time before deploy I will experiment, otherwise, I'll merge.

PR #319, opened 2017-01-20
Fixed various bugs with range request implementation

### LINKED ISSUE #318: Bug with media range requests: content-length is returning negative values
This bug is affecting production systems.

Visible effects include the audio on the listen page not loading properly. The error seems to happen the first time the audio is loaded and loads fine after a page request. I ~~have worked out Chrome sends different range request headers~~ don't know why our request range code is different in the two different scenarios:

| Action | Range request header | Content-Length | Content-Range | Response |
|--------|--------------------------|------------------|------------------|------------|
| Next button click | `Range:bytes=0-`  | `Content-Length:512001`  |`Content-Range:bytes 0-512000/591513`  | `206 Partial Content` |
|                             | `Range:bytes=512001-` | `Content-Length:20` | (empty) | `502 Bad Gateway` |
| Same page refresh | `Range:bytes=0-`  | `Content-Length:591513` | `Content-Range:bytes 0-591512/591513` | `206 Partial Content` |

I can't determine why the range-request behaviour is different for the different requests.

The response from the server is a `502 Bad Gateway` response. After intense debugging, it seems some part of the ruby codebase is returning a negative content-length for the second response:
```
Content-Range: bytes 512001-79511/591513
Content-Length: -432489
```

*Importantly*: The rails application sends these response headers back without throwing an exception. Passenger freaks out and [this case](https://github.com/phusion/passenger/blob/release-5.1.1/src/agent/Core/Controller/ForwardResponse.cpp#L119) responds with a generic bad gateway response that masks the error. The only way I found to get the above header information was to change `passenger_log_level` to level `7` (that level may not be required, but it needs to be greater than `3`).

# Tasks

- Isolate a test case against our range request code
- Patch the bug
- Investigate throwing exceptions in our code so these sorts of header based errors generate email notifications in the future

atruskie:
Closes #318

- Additionally adds a header verfication checker for Content-Length to
ApplicationController
- Fixes multiple bugs in range request

@cofiem pretty please review - this patch did my head in. There's beer in it for you 😉 🍺


spec/models/range_request_spec.rb
```diff
@@ -116,23 +137,105 @@
...
     expect(info[:range_start_bytes][4]).to eq(50)
-    expect(info[:range_end_bytes][4]).to eq(range_request.max_range_size + 50)
+    expect(info[:range_end_bytes][4]).to eq(range_request.max_range_size + 50 - 1)
+  end
 
-    expect(info[:range_start_bytes][1]).to eq(0)
-    expect(info[:range_end_bytes][1]).to eq(range_request.max_range_size)
+  context 'special open end range case' do
+    # this test case comes from a real-world production bug: https://github.com/QutBioacoustics/baw-server/issues/318
+    # the second part of a large range request triggers a negative content length and negative gradient in the content
```
cofiem: What does 'negative gradient' mean?

app/models/range_request.rb
```diff
@@ -368,31 +378,43 @@ def response_range(info)
...
         end_range = info[:range_end_bytes_max]
 
       end
 
       start_range = info[:range_start_bytes_min] if start_range < info[:range_start_bytes_min]
       end_range = info[:range_end_bytes_max] if end_range > info[:range_end_bytes_max]
-      end_range = start_range + @max_range_size if (end_range - start_range) > @max_range_size
+      # e.g. bytes=0-499, max_range_size=500 => 499 - 0 + 1 = 500 > 500
+      if (end_range - start_range + CONVERT_INDEX_TO_LENGTH) > @max_range_size
+        raise StandardError.new("Range request maximum exceeded")
```
cofiem: replace this with 
fail CustomErrors::BadRequestError, 'The requested range exceeded the maximum allowed.'

spec/models/range_request_spec.rb
```diff
@@ -116,23 +137,105 @@
     expect(info[:is_multipart]).to be_truthy
 
     expect(info[:range_start_bytes][0]).to eq(0)
-    expect(info[:range_end_bytes][0]).to eq(range_request.max_range_size)
```
cofiem: Oh, I see,  I compared the range start/end value to the length - which is off by one, since the value is a zero-based, inclusive index. Makes sense.

Just to confirm: if `max_range_size` is 10, and `range_end_bytes` is as high as it can be, then `range_end_bytes` would be 9? So need to subtract one from the max length of 10.

spec/models/range_request_spec.rb
```diff
@@ -116,23 +137,105 @@
...
+    # this test case comes from a real-world production bug: https://github.com/QutBioacoustics/baw-server/issues/318
+    # the second part of a large range request triggers a negative content length and negative gradient in the content
+    # range header.
+
+
+    # before bug fix:
+    # file_size:                                  822281
+    # request:                      "Range: bytes 512001-"
+    # info[:range_start]:                         512001
+    # info[:range_end]:                           310279       <-- problem, negative range!
```
cofiem: well, to be clear, the issue with the `:range_end` is that it is less than the `:range_start`

app/controllers/application_controller.rb
```diff
@@ -565,4 +568,13 @@ def should_check_authorization?
     !is_devise_controller && !is_admin_controller
   end
 
+  def validate_headers
```
cofiem: I like this! Nice idea to pick up any future issues.

spec/models/range_request_spec.rb
atruskie: As per https://tools.ietf.org/html/rfc7233#section-1.2 ... yes

spec/models/range_request_spec.rb
atruskie: Meaning point 2 is less than point 1. A bad phrasing on my part

Issue #292, opened 2016-06-08
deleter_id not populated when AudioEvents are archived


atruskie:
Deleted AudioEvents correctly have their `deleted_at` field set but not their `deleter_id`.

Two recent AudioEvents that were created and then deleted:

| audio_recording_id | start_time_seconds | end_time_seconds | low_frequency_hertz | high_frequency_hertz | is_reference | creator_id | updater_id | deleter_id | deleted_at                   | created_at                   | updated_at                   |
|--------------------|--------------------|------------------|---------------------|----------------------|--------------|------------|------------|------------|------------------------------|------------------------------|------------------------------|
| 346629             | 8.3127             | 19.2261          | 1894.9219           | 6804.4922            | f            | 9          | 9          |            | "2016-06-08 03:49:54.620177" | "2016-06-08 03:49:10.420445" | "2016-06-08 03:49:11.306575" |
| 347102             | 4.3886             | 16.0914          | 2411.7188           | 8441.0156            | f            | 1          | 1          |            | "2016-06-08 03:45:54.829618" | "2016-06-08 03:45:21.345739" | "2016-06-08 03:45:21.853717" |

It appears that `deleter_id` was correctly populated for about 4 months (July to November) in 2014. 588 audio events that are deleted have `deleter_id` set out of all of the 4534 deleted events (~13%).

Things to do:
- ensure unit tests check the presence of `deleter_id` after an event is deleted
- apply unit test checks changes to any other `acts_as_paranoid` model
- fix the AudioEvent `deleter_id` bug
- fix any other `deleter_id` bugs if they exist

PR #321, opened 2017-02-08
[WIP] Add cancel account button to Admin user edit view

### LINKED ISSUE #295: Add cancel account button to Admin user edit view
Allow admins to delete old accounts.

https://github.com/QutBioacoustics/baw-server/blob/develop/app/views/devise/registrations/edit.html.haml#L19

https://github.com/QutBioacoustics/baw-server/blob/develop/app/views/user_accounts/edit.html.haml

cofiem:
Closes #295


config/settings/development.yml
```diff
@@ -10,7 +10,7 @@ development:
      database: baw_local_dev
      pool: 5
      username: postgres
-     password: ''
+     password: 'anothersecret938@'
```
atruskie: this should not have been changed - or at least checked in?

config/settings/test.yml
```diff
@@ -10,7 +10,7 @@ test:
     database: baw_local_test
     pool: 5
     username: postgres
-    password: ''
+    password: 'anothersecret938@'
```
atruskie: this should not have been changed - or at least not checked in?

db/structure.sql
```diff
@@ -2,15 +2,16 @@
...
+-- Dumped from database version 9.5.5
+-- Dumped by pg_dump version 9.5.5
 
 SET statement_timeout = 0;
 SET lock_timeout = 0;
 SET client_encoding = 'UTF8';
 SET standard_conforming_strings = on;
 SET check_function_bodies = false;
 SET client_min_messages = warning;
+SET row_security = off;
```
atruskie: this should be reverted - it interferes with the postgres version we are using

doc/guides/guides.yml
```diff
@@ -1,3 +0,0 @@
----
```
atruskie: why was this deleted?

The stub was needed at one point

atruskie:
In addition to the review, I have a few questions (because this appears to be more than skin-deep changes):

User accounts are not archivable/paranoid:

1. Are they permanently deleted?
1. What happens to associated records?
  - I did not see a unit test that ensured all relevant records are deleted?
  - Not a question: this potential for exponential deletions scares me... we've got referential integrity enabled... what happens when the user is a modifier on another user's project/site/audio_event?
1. What happens to on-disk audio recordings, and other non-db stuff if an account is deleted


I'm tempted to suggest that users should be paranoid...

cofiem:
So, regarding deleting a user, two (and a half) main options:
1. **Delete** user completely. This requires changing all the linked items in the db to:
    - have fields that link to users be nullable
    - be set to another user
    - also be deleted when a user is deleted
2. **Archive** user. This means the user might still appear in some places (e.g. last modified by),
but the user cannot log in, and the user details are deleted.
3. Some **combination** of the two - some items are also deleted, some are kept.

Deleting is the most obvious/easiest to explain to users. It is also the most difficult to implement,
as users are associated with almost every other table in the db. Changing to nullable or setting to e.g.
`admin` or `nobody` or something looses information and might introduce other issues.

Archiving makes it easier on the developers, but is more complicated to explain to users.
Yes, we can delete preferences and other user-specific info (e.g. email?, user name?), 
but the users will still be in the db. Can they re-activate their account? What will be shown in the
UI when a 'deleted' user is the owner? Last modifier?

A combination might make the most sense, but is also even more complex - need to modify some tables
to be nullable/set to other user as well as keep the user.

What shall we do?

atruskie:
Definitely archiving - and we'll keep it simple for now.

Use the language "deactivate my account" to imply we will keep the data. Add explaining terminology: 
> When your account is deactivated, you can no longer log in, or make any changes. No other user will be able to see your identifiable information.

- can they reactivate their account? yes - but through a contact us form (admin intervention only)
- we'll keep identifiable information for now, and only anonymize (delete email/username) on request (TODO: add note to terms and conditions)
- we'll show user ids (e.g. in data exports and API responses) where relevant and in other places (e.g. the modified/created tags) we'll add some generic fallback text (maybe: "Anonymous" or "Deactivated user")

I think if we ensure the account is deactivated, we can fix up small UI bugs later on, so don't worry about all the edge cases so much.

(which is all much better than a hard deletion option because we'd have to consider every edge case).

cofiem:
Will likely close #292

PR #372, opened 2018-09-17
Feature progress events endpoints

### LINKED ISSUE #348: Implement Progress API
Implement the following endpoints for the new progress models. 

- Controller: `ProgressEvents`
  - `/datasets/{dataset_id}/progress_events` (`GET` & `:index`, `POST` & `:create`)
  - `/datasets/{dataset_id}/progress_events/filter`  (`GET` & `:filter`)
  - `/datasets/{dataset_id}/progress_events/{progress_events_id}`  (`GET` & `:show`, `DELETE` & `:destroy`)
- Controller: `ProgressEventsSummary`
  - `/datasets/{dataset_id}/progress_events/summary` (`GET` & `:index`, `POST & `:filter`)  

Note: we will not support updating progress events

---

- [ ] Make controllers for ProgressEvents and ProgressEventsSummary items (can use rails console)
- [x] Ensure controller definitions match the idioms found in our existing controllers
- [x] Add acceptance specs to cover all routes and user combinations
- [x] Add appropriate documentation to our Wiki about these new endpoints
- [ ] Ensure creating new progress events for the `Default` dataset works automatically

### LINKED ISSUE #368: Incorrect default order for dataset items
When requesting a list of dataset items without specifying the order, the default order should be by the "order" column, but instead it is sorted by audio_recording_id then order. 

In other words, requesting /datasets/$dataset_id/items/ incorrectly returns the same results as 
```SQL
select * from dataset_items
where dataset_id = $dataset_id
order by "audio_recording_id", "order"
```

### LINKED ISSUE #371: Alternative route for progress events that does not require a dataset_item_id
The Default dataset is used in conjunction with progress events to track user history on the listen page. Each time a user interacts with a 30 second section of audio, a progress event will be created, which will be associated with a dataset item in the default dataset. 

The dataset_item in the default dataset may already exist if any user has already visited the relevant segment of audio or may not exist if it's the first time anyone has visited that segment of audio. 

Rather than have the client first check if the dataset item exists, create the dataset item if needed, then create the progress event, this new route will create the dataset_item and the progress event at the same time. This avoids unnecessary requests between the client and server and reduces the complexity of the client. 

The dedicated route must therefore pass all the parameters needed to form a superkey of a DatasetItem (i.e. a set of attributes that uniquely identifies a dataset item). For dataset_items these are
- dataset_id
- audio_recording_id
- start_offset_seconds
- end_offset_seconds

This route is `POST /datasets/:dataset_id/progress_events/audio_recordings/:audio_recording_id/start/:start_offset_seconds/end/:end_offset_seconds`

The parameters of the progress event are included in the POST body. There is only one parameter:
- activity

This route also can be used wherever it is more convenient to use the set of identifying attributes rather than the id. 

In the case of progress tracking for the default dataset, the dataset_item may need to be created. This should be done by the action handling this route. Dataset items should not be created for other datasets. 

So, the steps performed by the action should be 
1. get the dataset_item specified by the parameters
    1. if it doesn't exist and the dataset_id is that of the default dataset, create the dataset item
    2. if it doesn't exist and the dataset_id is not the default dataset, respond with ... ? (maybe 404 or 400?)
2. with that dataset_item's id, create the progress event.

peichins:
Closes #371 
Fixes #368
(Included filter specs that return datasets that have different audio_recording_ids in the correct order)
Closes #348

Progress events api endpoints. The standard endpoints plus:
- a specific endpoint in dataset items controller #filter_todo that performs the filter with a priority algorithm that returns dataset items in order of their count of 'viewed' progress events.  It was named 'filter_todo' because it returns items in a dataset that are still "to do", i.e. they have not yet been viewed. I am happy for this to be replaced with a better name. 
- a specific endpoint for adding progress events with route parameters that define the parent dataset_item by its dataset_id, audio_recording_id, start_time_seconds, and end_time_seconds. This endpoint allows adding of progress events to segments of audio without knowing the dataset_item_id, or, in the case of the default dataset, if the dataset item has not yet been created.

### DIFF
--- a/app/controllers/progress_events_controller.rb
+++ b/app/controllers/progress_events_controller.rb
@@ -66,6 +66,8 @@ class ProgressEventsController < ApplicationController
 
     # find dataset_item
 
+
+
     dataset_item_params = {
         dataset_id: params[:dataset_id].to_i,
         audio_recording_id: params[:audio_recording_id],
@@ -73,12 +75,16 @@ class ProgressEventsController < ApplicationController
         end_time_seconds: params[:end_time_seconds]
     }
 
+    if params['dataset_id'] == 'default'
+      dataset_item_params[:dataset_id] = Dataset.default_dataset_id
+    end
+
     dataset_item = DatasetItem.find_by(dataset_item_params)
 
     resource_params = progress_event_params
     if dataset_item
       resource_params[:dataset_item_id] = dataset_item.id
-    elsif dataset_item_params[:dataset_id].to_i == 1
+    elsif params['dataset_id'] == 'default'
 
       # is for the default dataset, so create the dataset item
       dataset_item = DatasetItem.new(dataset_item_params)
--- a/app/controllers/progress_events_controller.rb
+++ b/app/controllers/progress_events_controller.rb
@@ -59,57 +59,22 @@ class ProgressEventsController < ApplicationController
     end
   end
 
+
   # POST /datasets/:dataset_id/progress_events/audio_recordings/:audio_recording_id/start/:start_time_seconds/end/:end_time_seconds
   # create a progress event without specifying a dataset item id by specifying the datset, audio_recording and offsets
   # will create an item in the default dataset if it does not exist
   def create_by_dataset_item_params
 
-    # find dataset_item
-
-
-
-    dataset_item_params = {
-        dataset_id: params[:dataset_id].to_i,
-        audio_recording_id: params[:audio_recording_id],
-        start_time_seconds: params[:start_time_seconds],
-        end_time_seconds: params[:end_time_seconds]
-    }
-
-    if params['dataset_id'] == 'default'
-      dataset_item_params[:dataset_id] = Dataset.default_dataset_id
-    end
-
-    dataset_item = DatasetItem.find_by(dataset_item_params)
-
     resource_params = progress_event_params
-    if dataset_item
-      resource_params[:dataset_item_id] = dataset_item.id
-    elsif params['dataset_id'] == 'default'
+    dataset_item = find_or_create_dataset_item
+    resource_params[:dataset_item_id] = dataset_item.id
 
-      # is for the default dataset, so create the dataset item
-      dataset_item = DatasetItem.new(dataset_item_params)
-      do_authorize_instance :create, dataset_item
-
-      if !dataset_item.save
-        fail CustomErrors::UnprocessableEntityError.new(
-            'Can not add progress event. Dataset item parameters were invalid.',
-            dataset_item.errors.messages
-        )
-      end
-
-      resource_params[:dataset_item_id] = dataset_item.id
-
-    else
-      fail CustomErrors::UnprocessableEntityError.new(
-          'Can not add progress event. Dataset item not found.',
-          dataset_item_params
-      )
-    end
-
+    # create a progress event instance, set its attributes and authorize it
     do_new_resource
     do_set_attributes(resource_params)
     do_authorize_instance :create
 
+    # finally, save the progress event
     if @progress_event.save
       respond_create_success(progress_event_path(@progress_event))
     else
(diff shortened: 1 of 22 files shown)


lib/modules/api/response.rb
```diff
@@ -282,6 +282,12 @@ def add_paging_and_sorting(new_query, filter_settings, filter_query)
         # execute a count against entire set without paging
         total = new_query.size
 
+        # if query involves aggregation, size returns the size of each group.
+        # and what we need is the number of groups
+        if total.is_a? Hash
```
atruskie: this case seems to be exclusive with the `.size` query above. as such the should be in an `elseif`

lib/modules/filter/query.rb
```diff
@@ -308,11 +308,17 @@ def apply_sort(query, table, column_name, allowed, direction)
       sort_field = @build.build_custom_field(column_name)
       sort_field = table[column_name] if sort_field.blank?
 
-      if direction == :desc
-        sort_field_by = Arel::Nodes::Descending.new(sort_field)
+      if (sort_field.is_a? String)
+        sort_field_by = sort_field
```
atruskie: does this need to be checked for sql injection?

I can't remember but I'm hoping the `params` methods do this in rails. Either that or by using it with `.order` below should ensure safety. Worth checking.

spec/acceptance/dataset_items_spec.rb
```diff
@@ -892,4 +893,104 @@ def body_params
...
+      standard_request_options(
+          :get,
+          'FILTER TODO (as invalid token)',
+          :unauthorized,
+          {expected_json_path: get_json_error_path(:sign_up)}
+      )
+    end
+
+    # not logged in users can filter dataset items, but they won't get any items that they don't have permission for
+    get '/datasets/:dataset_id/dataset_items/filter_todo' do
```
atruskie: is it worth adding a test case for public projects? I.e. a non-logged in user may have access to `audio_recordings` and `dataset_items`

spec/acceptance/progress_events_spec.rb
```diff
@@ -0,0 +1,918 @@
...
+        {response_body_content: ['"data":null'], expected_json_path: get_json_error_path(:permissions)}
+    )
+  end
+
+
+  ################################
+  # CREATE
+  ################################
+
+  # any user can create a dataset item if they have read-access on the project
```
atruskie: ~~this comment seems erroneous?~~

~~Only for the default dataset right?~~ 

Oh, I see, this is a copy-pasta comment. Please update.

spec/requests/dataset_items_spec.rb
```diff
@@ -123,6 +122,159 @@
...
+        dataset_item = FactoryGirl.create(:dataset_item,
+                                          creator: admin_user,
+                                          dataset: dataset,
+                                          audio_recording: audio_recordings[d % 2],
+                                          start_time_seconds: d,
+                                          end_time_seconds: d+10,
+                                          order: my_rand.rand * 10)
+        dataset_item.save!
+
+        cur_data = {dataset_item: dataset_item, progress_events: [], progress_event_count: 0 }
```
atruskie: do not abbreviate variable names

spec/requests/dataset_items_spec.rb
```diff
@@ -123,6 +122,159 @@
...
+
+        for c in progress_event_creators do
+          progress_event = nil
+          if d % c[:view_every] == 0
+            progress_event = FactoryGirl.create(:progress_event, creator: c[:creator], dataset_item: dataset_item, activity: "viewed", created_at: "2017-01-01 12:34:56")
+            #has_saved = progress_event.save!
```
atruskie: why is the save commented?

this section is very dense. Help the reader out please

spec/requests/progress_events_spec.rb
```diff
@@ -0,0 +1,199 @@
...
+                                      audio_recording: audio_recording,
+                                      start_time_seconds: 23,
+                                      end_time_seconds: 33,
+                                      order: 55.5)
+    another_dataset_item
+  }
+
+  let(:progress_event_attributes_invalid_dataset_item_id) {
+    attributes = progress_event_attributes
+    attributes[:dataset_item_id] = 38921111
```
atruskie: why the big id?

spec/requests/progress_events_spec.rb
```diff
@@ -0,0 +1,199 @@
...
+    @env['HTTP_AUTHORIZATION'] = reader_token
+    @env['CONTENT_TYPE'] = "application/json"
+
+    @create_progress_event_url = "/progress_events"
+    @update_progress_event_url = "/progress_events/#{progress_event.id}"
+  end
+
+  # converts values of a hash to string
+  # not recursive: does not work on nested hashes
+  def values_to_string hash
```
atruskie: i would have thought rails would have a similar utility?

app/controllers/progress_events_controller.rb
```diff
@@ -1,4 +1,150 @@
...
+        dataset_id: params[:dataset_id].to_i,
+        audio_recording_id: params[:audio_recording_id],
+        start_time_seconds: params[:start_time_seconds],
+        end_time_seconds: params[:end_time_seconds]
+    }
+
+    dataset_item = DatasetItem.find_by(dataset_item_params)
+
+    resource_params = progress_event_params
+    if (dataset_item)
```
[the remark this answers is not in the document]
peichins: This is part of the logic for determining whether a dataset item exists and if no create it if using the default dataset.

app/models/ability.rb
```diff
@@ -464,7 +477,37 @@ def to_dataset_item(user, is_guest)
...
+
+    # anyone can create as long as they have read access on the ancestor project of the dataset item
+    can [:create], ProgressEvent do |progress_event|
+      check_model(progress_event)
+
+      begin
+        Access::Core.can_any?(user, :reader, progress_event.dataset_item.audio_recording.site.projects)
+      rescue
+        fail CustomErrors::UnprocessableEntityError.new('Invalid dataset item')
+        false
```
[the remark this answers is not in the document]
peichins: why is this false here? At the time I was not sure if the fail would return from the method, so this ensures that permission is denied if this is the case. 
can you document why this try catch is required? 
this try-catch catches all types of errors: that is not ok.
It was there because the deeply nested call to projects failed if dataset item was invalid and therefore didn't have an audio recording associated. I have refactored it to explicitly check for audio recording and removed the try-catch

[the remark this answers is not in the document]
atruskie: cool. AFAIK that false never would have had any value. Even if the exception had been handled, that line would have been skipped. Also if we are returning values early in the method, it's better to use an explicit `return` keyword to remove ambiguity.

> I have refactored it to explicitly check for audio recording and removed the try-catch
Even better.

PR #479, opened 2020-07-15
Rswag - adds the ability to generate documentation for the API

atruskie:
# Rswag - adds the ability to generate documentation for the API

Adds documentation generation framework to rspec tests that supports open API spec

## Changes

The big change is the new `api` specs folder.
Documentation specs will no longer tests different permissions.

## Problems

- current acceptance tests are broken - don't have the time for a full rewrite. Disable?
- permission tests need to be redefined in request specs (preferraably using a shared helper)
- only proof of conept has been completed for Projects
- tags need to be changed from encoding permissions to grouping related resources

## Visual Changes

A new route has been mounted at `/api-docs`

image

## Final Checklist

- [ x Assign labels if you have permission
- [ ] Link issues related to PR


swagger/v1/swagger.yaml
```diff
@@ -0,0 +1,413 @@
+---
```
Allcharles, line 1: Isn't this file supposed to be auto-generated, does it need to be committed to git?

bin/rails
```diff
@@ -8,13 +8,31 @@ require_relative '../config/boot'
...
-  def default_options
-    super.merge!(Host: '0.0.0.0')
+        if ENV['HOST'] && !ENV['BINDING']
+          ActiveSupport::Deprecation.warn(<<-MSG.squish)
+          Using the `HOST` environment variable to specify the IP is deprecated and will be removed in Rails 6.1.
+          Please use `BINDING` environment variable instead.
+          MSG
+
+          return ENV['HOST']
+        end
```
Allcharles, line 28: Is there a reason the old method is still supported considering we are upgrading to 6.1? Considering we are using a docker container this should be under our control no?

app/models/project.rb
```diff
@@ -104,6 +104,23 @@ def self.filter_settings
...
+        name: { type: 'string' },
+        description: { type: 'string' },
+        description_html: { type: 'string', readOnly: true },
+        notes: { type: 'object' },
+        creator_id: { type: 'integer', readOnly: true },
+        updater_id: { type: 'integer', readOnly: true },
+        site_ids: { type: 'array', items: { type: 'integer' } }
+      }
+    }.freeze
+  end
```
Allcharles, line 122: Is it not possible to generate this from the existing project model? If not, default values like `creator_id` should be extracted and standardized across models

README.md
```diff
@@ -173,7 +167,18 @@ Use this style guide as a reference: https://github.com/rubocop-hq/ruby-style-gu
 
 ## Documentation
 
+Generate APi documentation with:
```
Allcharles, line 170:
```suggestion
Generate API documentation with:
```

README.md
```diff
@@ -153,15 +153,9 @@ docker-compose run --service-ports    --use-aliases  workers bash
 ```
 
 ### Tests
-The tests are run using Guard, either:
+The tests are run using rspec, either:
```
Allcharles, line 156:
```suggestion
The tests are run using rspec:
```

spec/api/projects_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+    delete('delete project') do
+      response(204, 'successful') do
+        schema nil
+        run_test! do
+          assert_no_response
+        end
+      end
+    end
+  end
+end
```
Allcharles, line 94: Will this be the new set of acceptance tests, or will it only show successful API responses?

swagger/v1/swagger.yaml
atruskie: both?

It could be committed -  which would make it a great static resource.
However, the tests currently generate volatile test data which means the examples would change every time the test is run.

I think I'll git ignore it, but I wanted you to see the output.

bin/rails
atruskie: Oh yeah, the old monkey patch was no longer effective on rails 6.1. I just hadn't noticed until now.

atruskie: Oh and to clarify, this patch changes the default behaviour for the options for `rails s'. We could always force a behaviour by setting an ENV var in the docker host, but that would prevent and of the CLI switches from working if we wanted any different behaviour (say running two versions of the web server for testing).

Changing the default option seems like the least intrusive choice

app/models/project.rb
atruskie: Good point. https://github.com/QutEcoacoustics/baw-server/issues/480

spec/api/projects_spec.rb
atruskie: Ideally acceptance tests will be phased out. 

- tests for functionality will become request specs
   - some new helpers for security access testing need to be added
- tests for API documentation of schema conformance will become API docs

And yes, the OAS specification supports documenting errors as well.

I'll add the above notes to some documentation

PR #483, opened 2020-07-28
Adds an automatic changelog generator

atruskie:
Added instructions for creating releases

## Changes
Invocation is manual at the moment since I don't have a neat way to auto-generate version numbers.

Also deleteded old travis file and updated docs.

## Problems

Can't figure out a sane way to auto increment versions. For now, leaving the version in the VERSION file should make this easy to script later on.

## Visual Changes

N/A

## Final Checklist


HISTORY.md
```diff
@@ -0,0 +1,420 @@
+# Old Changelog
```
Allcharles, line 1: Is there a reason behind splitting the changelog into separate CHANGELOG.md and HISTORY.md files?

atruskie: The changelog generator I`m using completely overwrites the old changelog.
It has a feature where it can import and old changelog if it is placed in HISTORY.md

atruskie:
> Looks good. I'm surprised the ruby ecosystem doesn't have somewhere already set which defines the version of a project, similar to the package.json file in node.

They do for Gems (packages). The .gemspec has a version field. Most libs simply define a `VERSION` constant in the root module. But it doesn't seem there is a consistent method for rails. I kept it in settings simply because that is where it was before 🤷‍♂️

PR #493, opened 2020-09-08
Adds a CMS

### LINKED ISSUE #492: Add content management system
Currently to customize our site we have view partials that need to be implanted into the filesystem of each instance.

This is a complex and annoying build step that happens in deployment. Ideally, we should be deploying the same code to every instance and it should only present customization through settings or database values.

Thus, we want to add a CMS that can store the images and page fragments needed to customize the site. Ideally the CMS should be headless.

atruskie:
# Adds a CMS

Fixes #492

Up until now we were relying on customizing the files, particularly the view templates, for each instance of the acoustic workbench.
This created expensive and complex deployment workflows when all we really needed was to pull some markdown blobs out of the database.

Additionally, at some point the new client would need to be able to programmatically access these views.

This commit adds a pre-built CMS named Comfortable Mexican Sofa. We chose CMS because it is mountable within our current app, it uses our active storage configuration, and it worked with rails 6 (unlike so many other solutions).

While we maintain backward compatibility we've also adapted the old public views to render the CMS blobs.

Changes are also needed to various spec helpers to accomodate the different content type and API shape exposed by CMS.

## Changes

- Adds a CMS system
- Public controller views no longer render partials from disk and instead render CMS blobs
- Some changes to spec helpers

## Visual Changes
image

image

## Final Checklist


spec/requests/cors_spec.rb
```diff
@@ -0,0 +1,157 @@
...
+      'User-Agent' => 'Mozilla/5.0 (Windows NT 6.3; WOW64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/39.0.2171.95 Safari/537.36',
+      'Access-Control-Request-Headers' => 'accept, content-type',
+      'Accept' => '*/*',
+      'Referer' => 'http://localhost:8080/listen/234234?start=30&end=60',
+      'Accept-Encoding' => 'gzip, deflate, sdch',
+      'Accept-Language' => 'en-US,en;q=0.8',
+      'Content-Type' => 'text/plain'
+    }
+
+    it '1, has the expected CORS response' do
```
Allcharles, line 26: Test names are a little confusing, what is 1 and 2?

spec/requests/public_spec.rb
```diff
@@ -2,156 +2,184 @@
...
+    )
+  end
+
+  it 'should render the CMS page' do
+    get '/credits'
+
+    expect(response_body).to include(
+      <<~HTML
+        <h2 id="if-you-like-pia-coladas">If you like piña coladas</h2>
+        <p>**And getting caught in the **rain
```
Allcharles, line 35: No bolding permitted?

spec/requests/public_spec.rb
```diff
@@ -2,156 +2,184 @@
...
+        I’m not much into health food
+        I am into champagne
+        I’ve got to meet you by tomorrow noon
+        And cut through all this red tape
+        At a bar called O’Malley’s
+        Where we’ll plan our escape”</p>
+      HTML
+    )
   end
 end
```
Allcharles, line 185: A lot of these tests seem redundant, or could be simplified. Why not just have one suite of tests with a few edge cases (h1, strong, em, etc), and then run the suite for each route?

Allcharles (approved):
Not sure if this PR is ready yet because you haven't pinged me on it, so my review might be a bit early. I have a couple points of concern:

- Are the CMS pages completely static? Or will we add the ability to provide an input which performs slight modification to the page. Ie. Home page, the image should be randomly selected, or the how to harvest page which changes instructions depending on url
- We need to discuss a more scalable method of managing CMS routes, still not sure on a good method
- CMS pages should be extracted to .md files so that non-programmers can edit them (ideally with some way for them to view the output without setting up/running the server themselves)
- How will the CMS reference links in the workbench-client? Are we going to output a special symbol which is then interpreted client side, or just use hard links?
  - Same issue if we want to use angular components like slideshows
- Will the CMS provide custom styling for the content? Or just use workbench-client defaults like bootstrap?

Relates to https://github.com/QutEcoacoustics/workbench-client/issues/49

atruskie:
@Allcharles it was ready for review. To answer your questions:

> Are the CMS pages completely static? Or will we add the ability to provide an input which performs slight modification to the page. Ie. Home page, the image should be randomly selected, or the how to harvest page which changes instructions depending on url

No, they're dynamic. There is an extremely limited amount of templating available (for security reasons) that can be extended via the `CmsHelpers` module. 

For random image selection the old site used some JavaScript. The new one can too.

The harvest page probably doesn't need to be a CMS page, since the instructions are invariant to the instances (the same between them). If customization is needed, we can insert a harvest blob in between the invariant blobs.

> We need to discuss a more scalable method of managing CMS routes, still not sure on a good method

With this change there are "well-known" pages that a guaranteed to exist in an instance. These are:

- /cms
- /cms/ethics
- /cms/privacy
- /cms/data_upload
- /cms/credits

Those tokens can be hard coded in the client and be assumed to exist. See `cms_seeds.rb` and note it runs every time the database is migrated.

Implementing a full website with sub-pages was not the goal of this change set, though it is possible.  The JSON responses to the any of the `/cms` pages return the child pages. One could implement a fully dynamic site from this information.
We could also expose an index of all available pages, but again, that is not a requirement for the current changeset.

> CMS pages should be extracted to .md files so that non-programmers can edit them (ideally with some way for them to view the output without setting up/running the server themselves)

No. They're stored as markdown in the database. The CMS system has a website for editing these pages (WYSIWYG and all) which is currently only available to the admin user. IF we decide to allow more users access we can look at that down the line. But again the current requirement is to get the files out of the disk and into to the database.

If the `cms_seeds.rb` is causing confusion for you: that file is used to inject default values into the database after migration. It does not overwrite pages, only replaces them if missing. Given its nature, normal people won't edit `cms_seeds.rb`, and we will rarely do so either, so a code based solution is sufficient.

After the default values are inserted into an instance's database, they can then be customised to whatever that particular instance wants them to say. E.g. the index blob will be changed via the cms backend (`/admin/cms`) once per staging, Ecosounds, and A2O.

> How will the CMS reference links in the workbench-client? Are we going to output a special symbol which is then interpreted client side, or just use hard links?

I don't imagine this will be needed. This system is mainly for storing small blobs of text that vary between instances. So we're using the client's menus, headers, and footers.

If this is needed, we can URL generation helpers to `CmsHelpers` that generate links, or we can have angular template the blobs.

> Same issue if we want to use angular components like slideshows

If the slideshow is common to all instances, it will be implemented outside of the CMS and not be an issue.
If a slideshow is only present in one instance, it will be implemented inside of the CMS, and likely use whatever mechanism the author chooses - so not the client's problem.

> Will the CMS provide custom styling for the content? Or just use workbench-client defaults like bootstrap?

The CMS system currently has a default layout that allows for outputting JavaScript and CSS. However, currently, the default is for both to be empty.

It currently outputs very sparse HTML fragments - no `<html>`, no `<head>`, no `<body>` - designed to be embedded directly into the page. The container for these blobs on the client should be formatted automatically by Bootstrap (classes may need to be applied to the container)

spec/requests/public_spec.rb
atruskie: They're not testing markdown rendering (though they do that too).

They're testing the Rails views now render blobs from the CMS system, and not partials from disk. They also test that the right view is rendering the right CMS content (i.e. `credits.html.haml` should render the the CMS page `/cms/credits` and not `/cms/privacy`)

They're a compatibility shim until these views are deleted when the new client is ready.

spec/requests/public_spec.rb
atruskie: Ah, bad test. I'll Fix.

Bold doesn't work if goes `**word **nextword`, it needs to be `**word** nextword`.

spec/requests/cors_spec.rb
atruskie: this file was renamed, and not part of the change set.

I didn't write these tests but I believe the numbers are different examples of requests

db/cms_seeds/cms_seeds.rb
```diff
@@ -0,0 +1,176 @@
...
+        - [AngularJS](https://angularjs.org/)
+        - [D3.js](https://d3js.org/)
+
+        ## Platforms and services
+
+        - [Github](https://github.com)
+        - [QRISCloud](https://www.qriscloud.org.au/)
+        - [nectar](https://https://nectar.org.au/)
+      MARKDOWN
+      }
```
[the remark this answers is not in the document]
atruskie: There is - but see main comments for why it isn't

config/initializers/comfortable_mexican_sofa.rb
```diff
@@ -0,0 +1,128 @@
...
+        only: [:full_path, :label]
+      }
+    }
+  }
+end
+
+# Default credentials for ComfortableMexicanSofa::AccessControl::AdminAuthentication
+# YOU REALLY WANT TO CHANGE THIS BEFORE PUTTING YOUR SITE LIVE
+#ComfortableMexicanSofa::AccessControl::AdminAuthentication.username = 'username'
+#ComfortableMexicanSofa::AccessControl::AdminAuthentication.password = 'password'
```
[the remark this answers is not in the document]
atruskie: Auth is handled by the module below in the same file.

We have no need to use the auth system, we use ours.

app/views/public/credits.html.haml
```diff
@@ -1,59 +1,4 @@
...
-  %li
-    %a{href: 'https://travis-ci.org'} TravisCI
-
-- if partial_exists?('public/credits')
-  %h1
-    = Settings.organisation_names.parent_site_name
-    Credits
-  = render partial: 'public/credits'
\ No newline at end of file
+= raw (controller.render cms_page: '/credits').gsub(/<h1>.*<\/h1>/, '')
```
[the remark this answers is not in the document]
atruskie: Temporary hack. The view system currently sticks its own `<h1>` in and thus two headings (one from rails, one from CMS) is redundant. Since we're deprecated all the rails views, this is a quick, and dirty, wasy to remove the other `<h1>`.

app/controllers/application_controller.rb
```diff
@@ -656,10 +656,8 @@ def set_last_seen_at
   end
 
   def should_check_authorization?
-    is_devise_controller = devise_controller?
-    is_admin_controller = respond_to?(:admin_controller?) && admin_controller?
-
-    !is_devise_controller && !is_admin_controller
+    return false if devise_controller?
+    return false if respond_to?(:admin_controller?) && admin_controller?
   end
```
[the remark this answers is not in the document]
atruskie: Good catch

PR #498, opened 2020-10-09
Fixes cut speed for ffmpeg

atruskie:
Fixes https://github.com/QutEcoacoustics/baw-audio-tools/issues/11

It looks we had slow cutting speed that scaled linearly with seek time. For long files, cut speed was nearing 8 seconds!

Added benchmarks as tests which should prevent future regressions.

Also include various tweaks and simplifications to the workers initialization code (it was necessary for debugging).

## Changes
Also:
- ensures workers bootsnap uses the correct tmp folder
- Adds a larger test fixture needed to test these sorts of problems
- Ensures ffmpeg is never interactive
- Ensures ffmpeg will overwrite files in race conditions
- Ensures redis communicator actually uses namespace option from config file
- Adds a sanity check to the multi-logger formatter setter to see if we can track down a persistent issue of Resque overwriting our formatter. See check_resque_formatter.
- ensures fixture pathnames exist by default
- enabled LFS to support large fixtures

## Problems

Not 100% convinced cut segments are properly cut. I've manually checked. I might add a test later.

## Visual Changes
None, but have a pretty graph:

image

## Final Checklist


Gemfile
```diff
@@ -42,6 +42,8 @@ gem 'rswag-ui'
 # uri parsing and generation
 gem 'addressable'
 
+gem 'descriptive_statistics', :require => 'descriptive_statistics/safe'
```
Allcharles, line 45: I would like you to start leaving links to gem repos so they can be checked more easily. My assumption is this comes from [here?](https://github.com/thirtysixthspan/descriptive_statistics), in which case it hasn't been updated in over a year. Repo owner doesn't appear to be actively developing it anymore: https://github.com/thirtysixthspan/descriptive_statistics/issues/48#issuecomment-490995025

lib/gems/baw-audio-tools/lib/baw_audio_tools/audio_ffmpeg.rb
```diff
@@ -44,7 +44,12 @@ def modify_command(source, source_info, target, start_offset = nil, end_offset =
       cmd_channel = arg_channel(channel)
       codec_info = codec_calc(target)
 
-      audio_cmd = "#{@ffmpeg_executable} -i \"#{source}\" #{cmd_offsets} #{cmd_sample_rate} #{cmd_channel} #{codec_info[:codec]} \"#{codec_info[:target]}\""
+      # AT 2020: added -nostdin and -y to force ffmpeg to complete in race conditions, and to never look for console input.
+      #   Both are speculative fixes. Races shouldn't happen but we think there's a bug with the unique job system.
+      # AT 2020: moved command offsets so that they occur before -i. This apparently changes performance profile for cut+seek
+      #   from O(n) to O(1), which makes a big difference on longer files!
+      #   For a 2 hour file cut speed drop to < 3 seconds, from 8.25 seconds.
```
Allcharles, line 51: Might be an idea to link to this PR for better context

lib/gems/baw-workers/lib/baw_workers/config.rb
```diff
@@ -339,6 +333,8 @@ def check_resque_formatter
 
         warn 'WARNING: Resque overwrote the default formatter!'
 
+        # when this is no longer an issue remove exception in lib/gems/baw-workers/lib/baw_workers/multi_logger.rb#formatter=
+
```
Allcharles, line 337: Create an issue?

spec/lib/gems/baw_audio_tools/audio_benchmarks.rb
```diff
@@ -0,0 +1,70 @@
...
+
+    mean = DescriptiveStatistics.mean(real_times)
+    variance = DescriptiveStatistics.variance(real_times)
+    standard_deviation = DescriptiveStatistics.standard_deviation(real_times)
+
+    # This doesn't strictly test for a constant variable with noise.
+    # But, the bounds are narrow enough that the test failed previously when seek time in ffmpeg was linear
+    expect(mean).to be_within(0.2).of(0.64)
+    expect(variance).to be_within(0.1).of(0.02)
+    expect(standard_deviation).to be_within(0.1).of(0.03)
```
Allcharles, line 68: Magic numbers

Gemfile
Allcharles: On another thought, we should potentially do a check of all the gems to see which ones have become un-maintained, and list them for replacement later on

spec/requests/media/benchmarks.rb
```diff
@@ -0,0 +1,95 @@
...
+    }.to perform_under(4).sec.warmup(0)
+
+    expect_success
+    expect(response.content_type).to eq('audio/wav')
+    expect(response.content_length).to be_within(100).of(1_323_078)
+
+    expect(response.headers[MediaPoll::HEADER_KEY_RESPONSE_FROM]).to eq MediaPoll::HEADER_VALUE_RESPONSE_REMOTE
+    expect(response.headers[MediaPoll::HEADER_KEY_ELAPSED_TOTAL].to_f).to be_within(0.5).of(3.25)
+    expect(response.headers[MediaPoll::HEADER_KEY_ELAPSED_PROCESSING].to_f).to be_within(0.5).of(3.25)
+    expect(response.headers[MediaPoll::HEADER_KEY_ELAPSED_WAITING].to_f).to be_within(0.1).of(0.0)
```
Allcharles, line 54: Magic numbers

Allcharles (approved):
A couple things:
- Why are there three binary files in the repo? Shouldn't they be in LFS
- Why was the size of the binary files changed?
- Are the files which are created/modified by the tests tracked? If so, should they not be instead added to the gitignore?
- While i like the performance benchmarks, how certain are we that the machine they run on will be consistent in speed? I would like the metrics to be extracted in such a way that modifying them, to match the potentially changing performance of the CI, becomes simple as it may need to be adjusted in future

atruskie:
> A couple things:
> 
> * Why are there three binary files in the repo? Shouldn't they be in LFS

You mean these?
image

They are in LFS. They weren't before. The LFS install process converted them.

> * Why was the size of the binary files changed?

See above.

> * Are the files which are created/modified by the tests tracked? If so, should they not be instead added to the gitignore?

No, they're all written into `tmp` which is ignored. Almost all are cleaned up after their tests run as well.

> * While i like the performance benchmarks, how certain are we that the machine they run on will be consistent in speed? I would like the metrics to be extracted in such a way that modifying them, to match the potentially changing performance of the CI, becomes simple as it may need to be adjusted in future

These are performance benchmarks yes, but they're also regression tests. The thresholds set are very generous, and I intend to adjust them if needed. As long as they pass on CI and fulfill their purpose, then we're good. They don't need to be the same speed everywhere (and won't fail with slight variances).

We do have metrics (for requests/response times) and adding generation timing figures to metrics will be useful. However, metrics are not the right tool for this job. This performance benchmarks assert that the system has certain properties and should not be released if these properties are invalidated. This regression was introduced (I think) during a ffmpeg upgrade some years ago. If we had had tests like these then, it would not have gone unnoticed for so long.

spec/lib/gems/baw_audio_tools/audio_benchmarks.rb
atruskie: Disagree. 
These are not values that change the behavior of the code.
The numbers relate directly to the values they're tested against and are thus described by the code around them.
Adding a name to these values adds no more meaning to them.

spec/requests/media/benchmarks.rb
atruskie: See https://github.com/QutEcoacoustics/baw-server/pull/498#discussion_r502218917

Gemfile
atruskie: > I would like you to start leaving links to gem repos so they can be checked more easily. 
No. See https://rubygems.org/gems/

Maintenance doesn't really matter for a library like this. Still, I found a more maintained version and am using that.

PR #522, opened 2021-07-20
Move status endpoint to dedicated controller

atruskie:
# Refactor of status endpoint

Simplify the public controller by breaking up functionality.

## Changes

- status action moved to a dedicated controller
- a new API only controller was created

## Problems

Unit tests were run manully for changes.

## Visual Changes

N/A

## Final Checklist


spec/requests/public_spec.rb
```diff
@@ -26,7 +24,7 @@ def update_page(slug, text)
     )
   end
 
-  it 'should render the CMS page' do
+  it 'renders the CMS page' do
```
Allcharles, line 27: Do these tests write out which CMS page they are validating? Its also likely that this should be an array of pages and expected responses. As we add CMS pages, we don't want the tests to balloon out

spec/requests/status_spec.rb
```diff
@@ -0,0 +1,111 @@
...
+
+describe '/status.json' do
+  before do
+    BawWorkers::Config.original_audio_helper.possible_dirs.each { |path|
+      FileUtils.mkdir_p(path)
+    }
+  end
+
+  example 'everything ok' do
+    get '/status.json'
```
Allcharles, line 11: We don't have a test for `/status`, or is it an assumption that it'll work because the `/status.json` version works?

Allcharles: Ahhh, I just saw in status_routing_spec you test that both return the same thing

Allcharles (approved):
Looks good. When doing these changes, I would like to see a version of the expected output in the PR so I know what is going to be returned. Trying to figure things out from the code can be fairly tricky depending on the route.
Is this route in the swagger document? If not, it may be a simple time to add it as I never knew it existed (likely because I never needed to know)

atruskie:
>expected output in the PR 

```
{"status":"good","timed_out":false,"database":true,"redis":"PONG","storage":"2 audio recording storage directories available.","upload":"Alive"}
```
> Is this route in the swagger document?

No, it's not. On purpose because it does not conform to the rest of our API standard. It's used internally for monitoring tools.

Is it worth adding an endpoint that does not conform?

spec/requests/public_spec.rb
atruskie: These aren't new tests.

This test in particular just makes sure that markdown is rendered and accessible by an endpoint.

All the CMS requests go through a single controller, so if one request works as expected, then the rest will too

Allcharles:
> 
> > Is this route in the swagger document?
> 
> No, it's not. On purpose because it does not conform to the rest of our API standard. It's used internally for monitoring tools.
> 
> Is it worth adding an endpoint that does not conform?

If anything, that means it should be documented in more detail. The explanation should be that it is an internal monitoring tool and should not be used; plus a general idea of how/where it is used. This will make it easier to understand if/when the api should be removed or potentially changed in the future to be compatible.

atruskie:
@Allcharles  https://github.com/QutEcoacoustics/baw-server/commit/25c6ad5a10fce1919e9ec2c8af015b6d6bed7b9a

PR #524, opened 2021-07-20
Fixes stats page and adds stats endpoint

### LINKED ISSUE #413: Missing statistics route
# Missing Route

Associated Page: https://www.ecosounds.org/website_status
Expected Route: `/statistics`
Expected Method: `GET`
Expected Body: empty
Expected Action: Should provide general statistics about the website including the following details:

- Number of projects
- Number of annotations
- Number of available tags
- Number of sites
- Number of audio recordings
- Number of users
- Number of unique tags attached to annotations
- Number of tags attached to annotations
- Number of new annotations in last month
- Overall annotation duration
- Number of users online
- Overall audio recording file size
- Number of new audio recordings in the last month
- Overall audio duration
- Recent annotations
- Recent audio recordings

### LINKED ISSUE #523: website_status page broken
Due to raw SQL queries 

```
An ActiveRecord::UnknownAttributeReference occurred in public#website_status:

  Query method called with non-attribute argument(s): "COALESCE(audio_recordings.updated_at, audio_recordings.created_at) DESC"
  app/controllers/public_controller.rb:377:in `recent_audio_recordings'


  app/controllers/public_controller.rb:377:in `recent_audio_recordings'
  app/controllers/public_controller.rb:160:in `website_status'
  app/controllers/application_controller.rb:641:in `set_then_reset_user_stamper'
```

atruskie:
# Fixes stats page and adds stats endpoint

Fixes #523,  Closes #413

## Changes

Factored stats out into it's own controller for neatness.

Had to maintain backwards compatibility with old website_status view, but it should be easy enough to rip that out later.

Added some generic scopes to any model that implements the timestamps protocol.

Removed all raw SQL from the stats queries.

## Problems

N/A

## Visual Changes

N/A

## Final Checklist

### DIFF
--- a/app/controllers/stats_controller.rb
+++ b/app/controllers/stats_controller.rb
@@ -20,6 +20,8 @@ class StatsController < ApiController
 
   def self.fetch_stats(current_user)
     online_window = 2.hours.ago
+    recent_window = 1.month.ago
+    recent_limit = 10
     {
       summary: {
         users_online: User.recently_seen(online_window).count,
@@ -27,17 +29,17 @@ class StatsController < ApiController
         online_window_start: online_window,
         annotations_total: AudioEvent.count,
         annotations_total_duration: AudioEvent.total_duration_seconds,
-        annotations_recent: AudioEvent.recent_within(1.month.ago).count,
+        annotations_recent: AudioEvent.recent_within(recent_window).count,
         audio_recording_total: AudioRecording.count,
-        audio_recording_recent: AudioRecording.created_within(1.month.ago).count,
+        audio_recording_recent: AudioRecording.created_within(recent_window).count,
         audio_recording_total_duration: AudioRecording.total_duration_seconds,
         audio_recording_total_size: AudioRecording.total_data_bytes.to_i,
         tags_total: Tag.count,
         tags_applied_total: Tagging.count
       },
       recent: {
-        audio_recordings: Access::ByPermission.audio_recordings(current_user).most_recent(10),
-        audio_events: Access::ByPermission.audio_events(current_user).most_recent(10)
+        audio_recordings: Access::ByPermission.audio_recordings(current_user).most_recent(recent_limit),
+        audio_events: Access::ByPermission.audio_events(current_user).most_recent(recent_limit)
       }
     }
   end
--- a/app/controllers/stats_controller.rb
+++ b/app/controllers/stats_controller.rb
@@ -1,0 +1,44 @@
+# frozen_string_literal: true
+
+# Controller for the stats endpoint
+class StatsController < ApiController
+  skip_authorization_check only: [:index]
+
+  # GET /stats
+  # only returns json
+  def index
+    result = StatsController.fetch_stats(current_user)
+
+    # Simplify the response, by mutation. Show only ids from active record models.
+    result[:recent].transform_values! do |query|
+      query.pluck(:id)
+    end
+
+    built_response = Settings.api_response.build(:ok, result)
+    render json: built_response, status: :ok, layout: false
+  end
+
+  def self.fetch_stats(current_user)
+    online_window = 2.hours.ago
+    {
+      summary: {
+        users_online: User.recently_seen(online_window).count,
+        users_total: User.count,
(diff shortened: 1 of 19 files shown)


config/routes.rb
```diff
@@ -704,6 +704,7 @@
   # site status API
   get '/status/' => 'status#index', defaults: { format: 'json' }
   get '/website_status/' => 'public#website_status'
+  get '/stats/' => 'stats#index', defaults: { format: 'json' }
```
Allcharles, line 707: `/website_stats`?

spec/requests/permissions/status_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+              })
+            },
+    action: :index
+  }
+
+  ensures :admin, :owner, :writer, :reader, :no_access, :harvester, :anonymous, :invalid,
+          can: [custom_index],
+          cannot: [:create, :update, :destroy],
+          fails_with: :not_found
+end
```
Allcharles, line 33: Shouldn't this be in PR #522

Allcharles (approved):
I like the change, a lot cleaner of an implementation than what I was expecting to occur

app/controllers/stats_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+        audio_recording_total: AudioRecording.count,
+        audio_recording_recent: AudioRecording.created_within(1.month.ago).count,
+        audio_recording_total_duration: AudioRecording.total_duration_seconds,
+        audio_recording_total_size: AudioRecording.total_data_bytes.to_i,
+        tags_total: Tag.count,
+        tags_applied_total: Tagging.count
+      },
+      recent: {
+        audio_recordings: Access::ByPermission.audio_recordings(current_user).most_recent(10),
+        audio_events: Access::ByPermission.audio_events(current_user).most_recent(10)
```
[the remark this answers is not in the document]
atruskie: When the `user` is `nil` the query will still happen. Then the `ByPermission` module will only show public recordings.

This is tested in specs with the `create_anon_heierarchy` factory helper.

Why 10? Because that is what it was, and we're retaining backward compat with the view. Also, this is not a standard CRUD API endpoint. It returns a freeform structure for a virtual model for which there is only one item.

app/models/audio_event.rb
```diff
@@ -298,8 +301,8 @@ def self.csv_query(user, project, site, audio_recording, start_offset, end_offse
         tags_others_ids.as('other_tag_ids'),
         Arel::Nodes::SqlLiteral.new(
           "'#{url_base}" + 'listen/\'|| "audio_recordings"."id" || \'?start=\' || ' \
-                '(floor("audio_events"."start_time_seconds" / 30) * 30) || ' \
-                '\'&end=\' || ((floor("audio_events"."start_time_seconds" / 30) * 30) + 30)'
+                           '(floor("audio_events"."start_time_seconds" / 30) * 30) || ' \
+                           '\'&end=\' || ((floor("audio_events"."start_time_seconds" / 30) * 30) + 30)'
         )
             .as('listen_url'),
```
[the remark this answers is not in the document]
atruskie: All of `Arel` is internal only. Long convoluted history here, but it did not used to be, now it is, and now they're slowly fixing their "leaky abstraction" and opening up parts of it.

Arel is the low-level query building DSL for constructing SQL queries. Anything the ActiveRecord classes don't implement has to be raw SQL or Arel. And raw SQL just got turned into an exception-causing error in the latest rails version, hence the problem, and hence this PR.

config/routes.rb
atruskie: Nah, I hate that name.

Note the API routes are: `status` and `stats`, it's only the rails view route that is `website_status`.

Also: in the future there will be `/stats` endpoints added onto `projects`, `sites`, and `regions`. I think it is more consistent in the long run to use a shorter name.

spec/requests/permissions/status_spec.rb
atruskie: Correct, I tacked it on when I realised I left it out.

PR #563, opened 2022-03-15
Time of day feature

### LINKED ISSUE #562: Filter recordings by time of day
A few users are asking for the ability to filter audio recordings by the time of the day in which they were recorded. Ideally, it would be possible to filter by date by time of day, however an easier solution may be to just add some custom fields to the recordings filter route which achieves this result.

atruskie:
# Time of day feature

Adds support to the filter API to filter by time of day.

Fixes #562 

## Changes

Adds support for expressions in the filter API. See relevant wiki section: https://github.com/QutEcoacoustics/baw-server/wiki/API:-Filtering#expressions-experimental

## Problems

Currently only tests written as a proof of concept.

## Final Checklist

- [ ] Assign reviewers if you have permission
- [ ] Assign labels if you have permission
- [ ] Link issues related to PR
- [ ] Remove/Reduce warnings from edited files
- [ ] Unit tests have been added for changes


spec/requests/audio_recordings_spec.rb
```diff
@@ -142,4 +142,180 @@
...
+
+      post '/audio_recordings/filter', params: body, **api_with_body_headers(reader_token)
+
+      expect_success
+    end
+
+    def make_result(site_id:, day:, hour:)
+      a_hash_including({
+        site_id:,
+        recorded_date: Time.new(2021, 10, day, hour, 0, 0, '+10:00'),
```
Allcharles, line 316: Does this get converted back to UTC? I am under the impression (although i may be wrong) that recorded_date should be in UTC time, and `recorded_date_timezone` is used to convert

spec/requests/audio_recordings_spec.rb
```diff
@@ -142,4 +142,180 @@
...
+        make_result(site_id: perth.id, day: 2, hour: 16),
+        make_result(site_id: sydney.id, day: 3, hour: 18),
+        make_result(site_id: perth.id, day: 3, hour: 16),
+
+        # DST cross-over - now no recordings in Sydney start at 18:00
+        make_result(site_id: perth.id, day: 4, hour: 16),
+        make_result(site_id: perth.id, day: 5, hour: 16),
+        make_result(site_id: perth.id, day: 6, hour: 16)
+      ])
+    end
```
Allcharles, line 299: Depending on want, may be good to duplicate these tests for `recorded_end_date`

spec/requests/audio_recordings_spec.rb
atruskie: I haven't run the tests yet, but all the dates will be compared to by their instant (as in no matter what offset they're in two things that happened in the real world at the same time will be  equivalent).

For the tests I chose to work in Brisbane time over UTC just because it is easier to reason a timezone I'm more familiar with. But the tests would work exactly the same way with UTC, just the numbers would be, IMO, harder to understand.

PR #585, opened 2022-06-15
Upload host split

atruskie:
# Upload host split

Production networking is more complex than dev envirionment. Production hosts are load balanced for security reasons so the public host advertised needs to be different from the host used to adminstrate our upload service.

Also wraps harvest#update in a row-lock transaction.

Also makes the app more resiliant to missing directories on start up - this is necessary for when storage systems are in maintenance mode.

## Final Checklist


lib/gems/baw-app/lib/custom_types.rb
```diff
@@ -12,7 +12,15 @@ module CustomTypes
...
       p = ::Pathname.new(value).expand_path
-      p.mkpath
+      # Super important that this doesn't crash in production
+      # We want to be able to serve requests even if our storage is not available
+      # but the app will crash while booting
+      begin
+        p.mkpath
+      rescue Errno::EACCES => e
+        puts "ERROR: Could not create directory #{p}, #{e}"
+        raise e if BawApp.dev_or_test?
```
Allcharles, line 22: Does this cancel the harvest and just return "Could not create directory" for the API response?

atruskie: Nah it's nothing to do with the harvest

It's checks directories are present in the config file and makes them if it can. It was failing when our NFS storage was experiencing down time which resulted in the entire site crashing... We had previously made website resilient to the storage going offline, it would still serve requests, not hang, etc.... Without this change that work would be ineffective

Issue #289, opened 2016-05-09
Encode license and attribution information in project models


atruskie:
With the introduction of public permissions (https://github.com/QutBioacoustics/baw-server/pull/265) and other issues related to discoverability (https://github.com/QutBioacoustics/Ecoacoustics/issues/88) we need to start considering how public users of our data will use it. These concerns were raised by @JessCappadonna, @towsey, and @cofiem as well.

Thus add the following fields to projects:
- A license chooser, with urls to chosen license
  - need a short selection of relevant licences (complete lack of restrictions, all rights reserved (default), share/resuse with attribution/citation)
  - or allow a user to link to their own  
    - Creative Commons (https://creativecommons.org/choose/)
    - ANDS http://www.ands.org.au/working-with-data/enabling-data-reuse/licensing-for-reuse, 
    - Github's non-software licenses: http://choosealicense.com/non-software/
- An attribution field (markdown enabled)
  - Add a suggested template:
  
  ```
  This data was collected by <name of researcher/group> with support of <grant information>.
  
  If you use this data in your research please cite:
  Surname F., (20xx) A publication or something. _Journal of things_, etc...
  ```
  - add a site-wide attribution field that appears on every project for the software (config file value)

hudson-newey:
# Project attributions mockup

Users want the ability to assign SPDX licenses or custom licenses to projects.

Amended [project model shape](https://github.com/QutEcoacoustics/baw-server/blob/master/app/models/project.rb):

```ruby
# == Schema Information
#
# Table name: projects
#
#  id                      :integer          not null, primary key
#  allow_audio_upload      :boolean          default(FALSE)
#  allow_original_download :string
#  deleted_at              :datetime
#  description             :text
#  image_content_type      :string
#  image_file_name         :string
#  image_file_size         :bigint
#  image_updated_at        :datetime
#  name                    :string           not null
#  notes                   :text
#  urn                     :string
#  created_at              :datetime
#  updated_at              :datetime
#  creator_id              :integer          not null
#  deleter_id              :integer
#  updater_id              :integer
#  license                 :text             # NEW ADDITION
#
# Indexes
#
#  index_projects_on_creator_id  (creator_id)
#  index_projects_on_deleter_id  (deleter_id)
#  index_projects_on_updater_id  (updater_id)
#  projects_name_uidx            (name) UNIQUE
#
# Foreign Keys
#
#  projects_creator_id_fk  (creator_id => users.id)
[...]

PR #669, opened 2024-01-15
Add nullable license field to Projects model

hudson-newey:
# Add nullable license field to Projects model

Given that projects can be publically available for viewing, users have commonly associated usage rights and licensing information with the projects. However, with existing API capabilities, this information is usually encoded in the `description` field.

The addition of a specific `license` attribute on projects will allow uses to either: Reference an SPDX license code or insert a custom license as free form text.

## Changes

- Adds a license filed to project models

## Issues

Fixes: #289 

## Problems

None

## Visual Changes

None

## Final Checklist


atruskie (approved):
looks good.

Like we discussed, at some point in the future it might be nice to know programmatically which projects use an open license, server side. But for now, we don't have a clear use case for the added complexity and I anticipate we can do an uplift later when it is needed.

PR #700, opened 2025-01-14
Dev improvements

atruskie:
Add more user ids to groups. This is not ideal, but is one of the only ways to make the compose solution work with bind mounts and files from non UID 1000 users.

Attempted to use docker compose watch to replace rerun. The feature is not ready yet.

Simplified rerun script to use a manually set environment variable.

Changed sftpgo to use it's own databases. When it shared a database with rails they often stepped on each other's toes. It also helps the solution to boot without deadlocks: it used to be that upload needed a database, web creates the db, but web can't start because it depends on upload.

Removed .touch and the web health check. The .touch files were made by any container, which means workers could end up signalling healthy for the web container.... which was incorrect. Could not determine any reason for the health check to remain.

Split entrypoint into dev_start. That way dev maintenance tasks are only run: a) in development and b) on the web service.


provision/upload/config/sftpgo.yaml
```diff
@@ -55,7 +55,7 @@ sftpd:
 data_provider:
   driver: postgresql
   # from config/settings.default.yml
-  name: baw_local_development
+  #name:
```
hudson-newey, line 58: Question: Why is this commented out? Why not just remove the property of give it a semantic name?

hudson-newey (approved):

provision/upload/config/sftpgo.yaml
atruskie: To signal intentional omission. I'll add a comment to make it clearer

andrew-1234 (approved):

PR #713, opened 2025-02-05
Adds a Verification API

### LINKED ISSUE #705: Verification API endpoint
With the creation of the workbench clients verification grid, users want a way to persist their verification's to the baw-server database.

Persisting verification's to the baw-server database also opens new development such as:

- verification consensus (https://github.com/QutEcoacoustics/baw-server/issues/706)

Users should be able to

- Filter by confidence score
- Filter by provenance

### Example payload

-- TODO

### Example response

-- TODO

andrew-1234:
# Adds a Verification API

The purpose is to implement a verifications table and API.  The motivation is to allow verifications made on the baw-workbench verification grid be persisted to the baw-server database.  closes #705 

## Changes

Adds a `Verification`:
- migration
- model
- model: filter settings
- model: schema
- controller
- model factory

Modifies existing files to include verification:
- routes
- abilities
- factory bot helpers
- model relations 

Adds relevant `Verification` specs:
- model
- requests
- permissions
- api
- routing 

## Problems

- [ ] A complete test run is needed and underway to identify problems + fixes will be appended to this request as needed

## Final Checklist

### DIFF
--- a/app/controllers/verifications_controller.rb
+++ b/app/controllers/verifications_controller.rb
@@ -1,0 +1,106 @@
+# frozen_string_literal: true
+
+# VerificationsController
+class VerificationsController < ApplicationController
+  include Api::ControllerHelper
+
+  # GET /verifications
+  # GET /audio_recordings/:audio_recording_id/audio_events/:audio_event_id/verifications
+  def index
+    do_authorize_class
+    get_audio_event
+
+    @verifications, opts = Settings.api_response.response_advanced(
+      api_filter_params,
+      list_permissions,
+      Verification,
+      Verification.filter_settings
+    )
+    respond_index(opts)
+  end
+
+  # GET /verifications/:id
+  # GET /audio_recordings/:audio_recording_id/audio_events/:audio_event_id/verifications/:id
+  def show
+    do_load_resource
+    do_authorize_instance
+
+    respond_show
+  end
+
+  # GET /verifications/new
+  def new
+    do_new_resource
+    get_resource
+    do_set_attributes
+    do_authorize_instance
+
+    respond_new
+  end
+
+  # POST /verifications
+  def create
+    do_new_resource
+    do_set_attributes(verification_params)
+
+    do_authorize_instance
+
+    if @verification.save
+      respond_create_success(shallow_verification_url(@verification))
+    else
+      respond_change_fail
+    end
+  end
+
+  # PUT/PATCH /verifications/:id
+  def update
+    do_load_resource
+    do_authorize_instance
+
+    if @verification.update(verification_update_params)
+      respond_show
+    else
+      respond_change_fail
+    end
+  end
+
+  # Handled in Archivable
+  # DELETE /verifications/:id
+
+  # GET|POST /verifications/filter
+  # GET|POST /audio_recordings/:audio_recording_id/audio_events/:audio_event_id/verifications/filter
+  def filter
+    do_authorize_class
+    get_audio_event
+
+    filter_response, opts = Settings.api_response.response_advanced(
+      api_filter_params,
+      list_permissions,
+      Verification,
+      Verification.filter_settings
+    )
+    respond_filter(filter_response, opts)
+  end
+
+  private
+
+  def verification_params
+    params.require(:verification).permit(
+      :confirmed, :audio_event_id, :tag_id
+    )
+  end
+
+  def verification_update_params
+    params.require(:verification).permit(
+      :confirmed
+    )
+  end
+
+  def get_audio_event #rubocop:disable Naming/AccessorMethodName
+    @audio_event = AudioEvent.find(params[:audio_event_id]) if params&.key?(:audio_event_id)
+  end
+
+  def list_permissions
+    Access::ByPermission.audio_event_verifications(current_user, audio_event: @audio_event)
+  end
+end
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -110,6 +110,7 @@ class Ability
       to_tag(user, is_guest)
       to_tagging(user, is_guest)
       to_user(user, is_guest)
+      to_verification(user, is_guest)
 
       to_analysis(user, is_guest)
       to_media(user, is_guest)
@@ -717,10 +718,11 @@ class Ability
   end
 
   def to_user(user, is_guest)
-    # admin only: :index, :edit, :update
+    # admin only: :index, :edit
     # :edit and :update are the Admin interface for editing any user
     # normal users edit their profile using devise/registrations#edit
 
+    # users can :update their own attributes on the user model via api
     # users can only view their own:
     can [:projects, :sites, :bookmarks, :audio_events, :audio_event_comments, :update], User, id: user.id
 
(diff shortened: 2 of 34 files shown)


hudson-newey (approved):

app/models/ability.rb
```diff
@@ -825,4 +826,51 @@ def to_response(user, is_guest)
...
+    can [:create, :update], Verification do |verification|
+      check_model(verification)
+      Access::Core.check_orphan_site!(verification.audio_event.audio_recording.site)
+
+      has_writer_access = Access::Core.can_any?(
+        user, :writer,
+        verification.audio_event.audio_recording.site.projects
+      )
+
+      # only check creator_id for persisted records (updates)
```
atruskie, line 851:
```suggestion
      # anyone with writer access can create a verification, 
      # but only users who created a verifications can update them (and only if they still have access to the project)
```

app/models/tag.rb
```diff
@@ -33,6 +33,8 @@ class Tag < ApplicationRecord
   has_many :taggings, inverse_of: :tag
   has_many :audio_events, through: :taggings
   has_many :tag_groups, inverse_of: :tag
+  has_many :verifications, inverse_of: :tag
```
atruskie, line 36: missing dependent option

app/models/verification.rb
```diff
@@ -0,0 +1,138 @@
...
+#  fk_rails_...  (updater_id => users.id)
+#
+class Verification < ApplicationRecord
+  belongs_to :audio_event, inverse_of: :verifications
+  belongs_to :tag, inverse_of: :verifications
+  belongs_to :creator, class_name: 'User', inverse_of: :created_verifications
+  belongs_to :updater, class_name: 'User', inverse_of: :updated_verifications, optional: true
+
+  # A user can only have one verification per audio event and tag
+  validates :creator_id, uniqueness: { scope: [:audio_event_id, :tag_id] }
```
atruskie, line 40: if this is enforced a database constraint then this is doubling the work the database has to do.

it is not necessarily something to remove: it can be a nicer experience for rails validation errors to be reported and to also enforce integrity in the database, but we have to decide if it is worth the cost.

app/models/verification.rb
```diff
@@ -0,0 +1,138 @@
...
+  def self.filter_settings
+    fields = [
+      :id, :confirmed, :audio_event_id, :tag_id, :creator_id,
+      :updater_id, :created_at, :updated_at
+    ]
+
+    {
+      valid_fields: fields,
+      render_fields: fields,
+      text_fields: [:confirmed],
```
atruskie, line 83:
```suggestion
      text_fields: [],
```

text fields enable text based operations like `contains` - there's no point enabling those on enum columns. Don't need to run `contains` on a single word

app/models/verification.rb
```diff
@@ -0,0 +1,138 @@
...
+          on: Verification.arel_table[:audio_event_id].eq(AudioEvent.arel_table[:id]),
+          available: true
+        },
+        {
+          join: Tag,
+          on: Verification.arel_table[:tag_id].eq(Tag.arel_table[:id]),
+          available: true
+        }
+      ]
+    }
```
atruskie, line 109: probably need to expand these - it's quite conceivable we could get a filter query for verifications like "find me all verifications which are in project 123".

But let's keep the scope of this PR small, and add the associations when @hudson-newey complains

app/controllers/verifications_controller.rb
```diff
@@ -0,0 +1,92 @@
...
+      respond_change_fail
+    end
+  end
+
+  # PUT/PATCH /verifications/:id
+  def update
+    do_load_resource
+    do_authorize_instance
+
+    if @verification.update(verification_params)
```
atruskie, line 61: i'm wondering: do we even want to allow updating any field other than `confirmed`?

what is the point in allowing updated to `tag_id` or `audio_event_id`

config/routes.rb
```diff
@@ -747,6 +747,7 @@ def matches?(request)
       end
       resources :tags, only: [:index], defaults: { format: 'json' }
       resources :taggings, except: [:edit], defaults: { format: 'json' }
+      resources :verifications, only: [:index, :show], defaults: { format: 'json' }
```
atruskie, line 750: add `filterable`

db/migrate/20250120064731_create_verifications.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+class CreateVerifications < ActiveRecord::Migration[7.2]
+  def change
+    create_enum :confirmation, ['true', 'false', 'unsure', 'skip']
+
+    create_table :verifications do |t|
+      t.references :audio_event, null: false, foreign_key: { on_delete: :cascade }
+      t.references :tag, null: false, foreign_key: true
```
atruskie, line 9: missing cascade?

db/migrate/20250120064731_create_verifications.rb
```diff
@@ -0,0 +1,22 @@
...
+      t.column :creator_id, :integer, null: false
+      t.column :updater_id, :integer
+      t.column :confirmed, :confirmation, null: false
+
+      t.timestamps
+    end
+
+    add_foreign_key :verifications, :users, column: :creator_id
+    add_foreign_key :verifications, :users, column: :updater_id
```
atruskie, line 18: missing cascade?

spec/api/verifications_spec.rb
```diff
@@ -0,0 +1,141 @@
...
+
+    sends_json_and_expects_json
+    with_authorization
+    for_model Verification
+    which_has_schema ref(:verification)
+
+    let(:audio_recording_id) { audio_recording.id }
+    let(:audio_event_id) { audio_event.id }
+
+    path '/audio_recordings/{audio_recording_id}/audio_events/{audio_event_id}/verifications' do
```
atruskie, line 110: add filter spec

spec/models/verification_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+  end
+
+  it 'allows nil for updater_id' do
+    expect(build(:verification, updater_id: nil)).to be_valid
+  end
+
+  it 'allows updater_id to be set' do
+    v = create(:verification)
+    expect(build(:verification, updater_id: v.creator_id)).to be_valid
+  end
```
atruskie, line 74: generally don't need to test the user stamping methods

spec/permissions/verifications_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+    can_do: Set[:show, :index],
+    fails_with: :not_found
+
+  the_user :anonymous, can_do: Set[:index], fails_with: [:not_found, :unauthorized]
+
+  the_user :invalid, can_do: nothing, fails_with: [:not_found, :unauthorized]
+
+  the_user :no_access, can_do: Set[:index], fails_with: [:not_found, :forbidden]
+
+  the_user :harvester, can_do: nothing, fails_with: [:not_found, :forbidden]
```
atruskie, line 53: will need to change once filter is added

spec/requests/verifications_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+  before do
+    create(:verification, audio_event:, creator: owner_user, confirmed: 'true')
+    create(:verification, audio_event:, creator: writer_user, confirmed: 'false')
+  end
+
+  it 'a reader can list verifications' do
+    get '/verifications', **api_headers(reader_token)
+    expect(response).to have_http_status(:ok)
+    expect_at_least_one_item
+  end
```
atruskie, line 15: seems redundant? permission spec should cover this

spec/requests/verifications_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+    filter = {
+      filter: {
+        creator_id: { not_eq: owner_user.id }
+      }
+    }
+    get '/verifications/filter', params: filter, **api_headers(writer_token)
+
+    expect(response).to have_http_status(:ok)
+    expect_number_of_items(2)
+    expect(api_data).to include(a_hash_including(creator_id: writer_user.id))
```
atruskie, line 40: technically, this only tests that one of the two returned items has the writer user id

atruskie (changes requested):
Excellent work. You've paid close attention to the idiosyncrasies and idioms used in this solution.

Minor cleanup work required.

Note: this change set will not be sufficient enough to power the sort of advanced queries that @hudson-newey will need to power the verification interface (e.g. filter audio events where events have no verification, or filter audio events for event where verification consensus is less than 0.5, etc...) but we'll tackle those features in smaller future PRs.

atruskie:
We discussed offline: we should also change the values of the confirmed enum, specifically to avoid encoding primitive values ('true' and 'false') in the enum

app/controllers/verifications_controller.rb
andrew-1234: agree, updating other fields wouldn't makes sense

app/models/verification.rb
andrew-1234: Would that mean considering the front end implementation to understand the
utility here? Like, since adding a verification has to be unique
for the user / audio event / tag, choosing to verify something means that
information is set and the user can't manipulate them. So there is either an
existing verifcation for that combination or not, which changes the actions available to
show rather than create or something. In which case the rails verification
would be unnecessary?

atruskie: It's a good point that the validation won't be surfaced to the user of the website.

But it will be surfaced to consumers of the API. An error message with details is more useful than a generic 406 not acceptable.

Try it out! Remove the validation, write a request spec, see how ergonomic it is. Do we surface a useful error message? I can't remember!

andrew-1234: ActiveRecord::RecordNotUnique:
       PG::UniqueViolation: ERROR:  duplicate key value violates unique constraint "idx_on_audio_event_id_tag_id_creator_id_f944f25f20"
       DETAIL:  Key (audio_event_id, tag_id, creator_id)=(1, 1, 3) already exists.

atruskie (approved):
Well done 🥳

PR #718, opened 2025-02-18
Adds unique constraint to path on harvest_items

### LINKED ISSUE #660: Add a unique constraint to `path` on `harvest_items`
We found a case where a duplicate harvest_item row was made for the same file on disk. The `path` value was the same.

Consider adding a unique constraint to the harvest_items table.

Also does the btree index we're using not ensure uniqueness? Probably not. Seems like we missed something here.

Found by @hudson-newey

andrew-1234:
# Adds unique constraint to path on harvest_items

The purpose of this PR is to update the `path` field on harvest_items to a unique index constraint. This will address the issue of duplicate harvest_item rows (same file path) being persisted to the harvest_items table, which occurs due to a race condition during the audio file enqueue process of a harvest job.  

Closes #660

## Changes

- migration to add unique constraint on path and delete existing duplicates in harvest items table
- migration test 
- handle exceptions such that duplicate path doesn't cause db exception to be raised
  -  behaviour tested in `harvest_job_spec`
- add a helper for shared test logic (creating harvest_items with validations)
- fixes across tests where setup included creating harvest items with duplicate paths

## Problems

- Database backup prior to running migration
- Test suite full run in progress

## Final Checklist

- [ x] Assign reviewers if you have permission
- [ x] Assign labels if you have permission
- [ x] Link issues related to PR
- [ x] Remove/Reduce warnings from edited files
- [ x] Unit tests have been added for changes
- [  ] Ensure CI build is passing

### DIFF
--- a/spec/lib/gems/baw_workers/jobs/harvest/harvest_job_spec.rb
+++ b/spec/lib/gems/baw_workers/jobs/harvest/harvest_job_spec.rb
@@ -132,18 +132,44 @@ describe BawWorkers::Jobs::Harvest::HarvestJob, :clean_by_truncation do
       )
     end
 
-    def enqueue_and_perform(should_harvest:, completed: 1)
-      enqueued = BawWorkers::Jobs::Harvest::HarvestJob.enqueue_file(
+    def enqueue(should_harvest:)
+      BawWorkers::Jobs::Harvest::HarvestJob.enqueue_file(
         harvest,
         @paths.harvester_relative_path,
         should_harvest:
       )
+    end
 
+    def enqueue_and_perform(should_harvest:, completed: 1)
+      enqueued = enqueue(should_harvest: should_harvest)
       expect(enqueued).to be true
       expect_enqueued_jobs(1, of_class: BawWorkers::Jobs::Harvest::HarvestJob)
 
       perform_jobs(count: 1)
       expect_jobs_to_be completed:, of_class: BawWorkers::Jobs::Harvest::HarvestJob
+    end
+
+    context 'when simulating enqueue race condition' do
+      it 'database unique constraint is upheld gracefully' do
+        enqueue(should_harvest: true).tap do |enqueued|
+          expect(enqueued).to be true
+          expect(HarvestItem.count).to eq 1
+        end
+
+        # the race condition happens between the existence check and the record insertion
+        # so to simulate, we make the existence check return false, as if it would during a race
+        allow(BawWorkers::Jobs::Harvest::HarvestJob).to receive(
+          :existing_harvest_item
+        ).and_return(nil)
+
+        # before this test, an ::ActiveRecord::RecordNotUnique would have been thrown here
+        enqueue(should_harvest: true).tap do |enqueued|
+          expect(enqueued).to be false
+          expect(HarvestItem.count).to eq 1
+        end
+
+        clear_pending_jobs
+      end
     end
 
     context 'when getting metadata from a file' do
@@ -224,7 +250,7 @@ describe BawWorkers::Jobs::Harvest::HarvestJob, :clean_by_truncation do
         expect_delayed_jobs(1)
       end
 
-      it 'will delete the harvest file after a period of time' do
+      it 'deletes the harvest file after a period of time' do
         enqueue_and_perform(should_harvest: true)
 
         # @type [HarvestItem]
--- a/spec/migrations/add_path_unique_to_harvest_items_spec.rb
+++ b/spec/migrations/add_path_unique_to_harvest_items_spec.rb
@@ -1,0 +1,119 @@
+# frozen_string_literal: true
+
+require_migration!
+
+describe AddPathUniqueToHarvestItems, :migration do
+  let(:longer_blob) do
+    { error: 'null', fixes: [{ version: 1, problems: { FL001: { status: 'NoOperation' } } }] }
+  end
+  let(:datetime) { "'#{Time.zone.now}'" }
+  let(:harvest_items) { create_harvest_items }
+
+  before do
+    # format the harvest items into a string of values for insertion
+    harvest_values = harvest_items.map { |x| "(#{x})" }.join(",\n")
+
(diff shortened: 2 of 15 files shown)


spec/requests/harvest/harvest_batch_benchmark_spec.rb
```diff
@@ -76,20 +76,29 @@
...
+             values.map { |v| ActiveRecord::Base.connection.quote(v) }].join(', ')})"
+      }.join(', ')
+
+      ActiveRecord::Base.connection.execute(
+        <<~SQL.squish
+          INSERT INTO harvest_items
+          (path, info, audio_recording_id, #{columns})
+          VALUES #{values_list}
+        SQL
+      )
```
atruskie, line 101: maybe it's late, but i'm finding this really hard to parse.

My instinct again, is to revert your changes, and to proceed with a smaller changeset. 

Why does the performance matter? I doubt it is measurably faster.

Why is there so much extra generation?

Does the non-duplicated paths even need to exist?

spec/requests/harvest_items_spec.rb
```diff
@@ -154,10 +154,10 @@
 
   context 'with lots of items' do
     before do
-      50.times do
-        create_with_validations(fixable: 1, not_fixable: 0, sub_directories: 'a')
+      50.times do |_i|
```
atruskie, line 157: unused var

spec/support/harvest_item_helpers.rb
```diff
@@ -0,0 +1,53 @@
...
+
+# Helper methods for creating harvest items in tests
+module HarvestItemHelper
+  # Example methods for harvest item creation
+  module Example
+    # @return [Enumerator] sequence starting from 1
+    def sequence
+      @sequence ||= (1..).each
+    end
+
```
atruskie, line 11: yeah, this maintains state between tests? seems dangerous

spec/support/harvest_item_helpers.rb
```diff
@@ -0,0 +1,53 @@
...
+          status: :not_fixable,
+          name: :kiww_the_wabbit,
+          message: nil
+        )
+      end
+
+      info = BawWorkers::Jobs::Harvest::Info.new(
+        validations:
+      )
+      path = generate_recording_name(Time.zone.now, suffix: sequence.next.to_s)
```
atruskie, line 38: i'm wondering why this was even needed?

i'm guessing because the timestamp resolution was too coarse at the seconds level and that resulted in duplicates?

maybe just take the timestamp value, pull out the `.subsec.to_f` and use that as the suffix?

atruskie (changes requested):
Well done. Most of the suggestions are comments or style based.

spec/migrations/add_path_unique_to_harvest_items_spec.rb
```diff
@@ -0,0 +1,122 @@
...
+
+    ActiveRecord::Base.transaction do
+      ActiveRecord::Base.connection.execute(
+        <<~SQL.squish
+          DELETE FROM harvest_items;
+          INSERT INTO users (user_name, email, encrypted_password)
+          VALUES ( 'user1', 'user1@example.com', 'drowssap');
+
+          INSERT INTO projects (name, creator_id)
+          VALUES ('project1', #{select_one_from('users')});
```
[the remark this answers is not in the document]
andrew-1234: it's used on 37, harvest needs a not-null project

as far as i can tell a region isn't needed  in the mock data for the test to pass unless i've misunderstood something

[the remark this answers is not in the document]
atruskie: Ok yeah, you're correct. 

You have still created an orphaned site though. This will fail when we eventually remove projects_sites

spec/lib/gems/baw_workers/jobs/harvest/harvest_job_spec.rb
```diff
@@ -132,20 +132,47 @@
...
+
+        # the race condition happens between the existence check and the record insertion
+        # so to simulate, we make the existence check return false, as if it would during a race
+        allow(BawWorkers::Jobs::Harvest::HarvestJob).to receive(
+          :existing_harvest_item
+        ).and_return(nil)
+
+        # before this test, an ::ActiveRecord::RecordNotUnique would have been thrown here
+        enqueue(should_harvest: true).tap do |enqueued|
+        enqueue(should_harvest: true).tap do |enqueued|
```
atruskie, line 167:
```suggestion
        enqueue(should_harvest: true).tap do |enqueued|
```

spec/migrations/add_path_unique_to_harvest_items_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+  let(:datetime) { "'#{Time.zone.now}'" }
+  let(:harvest_items) { create_harvest_items }
+
+  before do
+    # format the harvest items into a string of values for insertion
+    harvest_values = ''
+    harvest_items.each.with_index(1) do |example, index|
+      delimit = index == harvest_items.length ? '' : ",\n"
+      harvest_values += "(#{example})#{delimit}"
+    end
```
atruskie, line 18: didn't like this suggestion? https://github.com/QutEcoacoustics/baw-server/pull/718#discussion_r1959215908

spec/requests/harvest/harvest_batch_benchmark_spec.rb
```diff
@@ -16,6 +16,7 @@
     let(:duplicates) { 1000 }
 
     before do
+      # debugger
```
atruskie, line 19: stray debugger

atruskie (approved):
Great work, a couple of final suggestions. Merge when ready.

spec/migrations/add_path_unique_to_harvest_items_spec.rb
andrew-1234: whoops, I missed that. i like it, will commit

PR #749, opened 2025-04-01
Filters out not-ready recordings from audio recording endpoints

### LINKED ISSUE #736: Ensure aborted recordings are filtered out of common scenarios
Audio recordings that are not `status: :ready` should not be emitted

- [ ] in index/filter scenarios (unless you're an admin?)
- [ ] used in analysis jobs

atruskie:
- also introduced the concept of filter defaults
- filter defaults can be overridden
- sets a filter default for status = ready for audio recordings filter settings

- also updates `filter_as_relation` for AnalysisJob to ensure only status ready recordings are used (not implemented as a default filter because we don't want it to be overridable)

Fixes #736

### DIFF
--- a/app/modules/filter/query.rb
+++ b/app/modules/filter/query.rb
@@ -29,6 +29,10 @@ module Filter
       :qsp_generic_filters, :paging, :sorting, :build,
       :custom_fields2, :capabilities
 
+    # Returns the filter provided by a request, before being merged onto the default filter
+    # @return [Hash] filter
+    attr_reader :supplied_filter
+
     # Convert a json POST body to an arel query.
     # @param [Hash] parameters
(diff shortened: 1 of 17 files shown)


app/modules/filter/query.rb
```diff
@@ -310,6 +344,60 @@ def add_qsp_to_filter(filter, additional, combiner)
       filter
     end
 
+    # Merge default filter into supplied filter.
+    # @param [Hash,Array] filter
+    # @param [Hash,Proc,nil] default
+    # # @param [Boolean] was_normalized
```
andrew-1234, line 350:
```suggestion
    # @param [Boolean] was_normalized
```

andrew-1234 (approved):

PR #758, opened 2025-05-01
Event summary report WIP

### LINKED ISSUE #653: New graph generation routes
As part of the audio event summary reports, the client needs information to create the accumulation Data plot, species composition plot, analysis coverage plots, and false colour spectrograms.

---

```jsonc
"accumulationData": [
    { "date": "2020-01-03T00:00:00.000Z", "countOfSpecies": 1, "error": 0.5 },
    { "date": "2020-01-04T00:00:00.000Z", "countOfSpecies": 2, "error": 0.5 }
],
```

image

---

```jsonc
"speciesCompositionData": [
	// each of these object data points will be of a bin size
    {
    	"date": "2020-01-03T00:00:00.000Z",
        "values": [
        	{ "tagId": 2, "ratio": 0.7 },
            { "tagId": 1, "ratio": 0.3 }
    	]
	},
	{
    	"date": "2020-01-04T00:00:00.000Z",
        "values": [
        	{ "tagId": 1, "ratio": 0.5 },
            { "tagId": 4, "ratio": 0.5 }
            ]
		}
        // etc...
],
```

image

---

False colour spectrograms should be returned as a url location or a collection of url locations.
A collection is an option because false colour spectrograms can only be created for 24 hour periods, and therefore, should return a collection or sorted url to false colour spectrograms
```jsonc
"falseColorSpectrograms": [ "https://api.ecosounds.org/false_color123.png" ],
```

image

We should either
- a: Expose each of these graphs under different atomic routes
- b: Expose these graphs under a common route (e.g. `/{project,region,site}/:id/audio_events/graphs` and have filter conditions)

andrew-1234:
Draft pull request for audio_event_report

Fixes: #653


app/modules/report/audio_event_report.rb
```diff
@@ -0,0 +1,552 @@
...
+      # tables without indexing on the cte array and using `.left`
+      accumulation_series_ctes, accumulation_series_aggregate = TimeSeries::Accumulation.accumulation_series_result(
+        base_table, @parameters
+      )
+
+      event_summary_ctes, event_summaries_aggregate = event_summary_result(base_table)
+
+      # hack. access the table using .left
+      bucketed_time_series = accumulation_series_ctes.find { _1.left.name == 'bucketed_time_series' }
+      base_verification = event_summary_ctes.find { _1.left.name == 'verification_base' }
```
atruskie, line 26: magic string

app/modules/report/audio_event_report.rb
```diff
@@ -0,0 +1,552 @@
...
+      'audio_recordings.recorded_date + CAST(audio_events.start_time_seconds || \' seconds\' as interval) ' \
+        'as start_time_absolute'
+    end
+
+    # returns an expression that projects the absolute end time of an audio
+    # event as a derived columnn
+    def end_time_absolute_expression
+      'audio_recordings.recorded_date + CAST(audio_events.end_time_seconds || \' seconds\' as interval) ' \
+        'as end_time_absolute'
+    end
```
atruskie, line 429: pretty sure these already exist elsewhere in the solution

helpers like this can go on the AudioEvent model

app/modules/report/time_series/coverage.rb
```diff
@@ -0,0 +1,405 @@
...
+          'make_interval',
+          [Arel.sql('secs => ?', seconds)]
+        )
+      end
+
+      # returns an expression that projects the absolute end date of an audio
+      # event as a derived columnn
+      def arel_recorded_end_date
+        # AudioRecording.arel_table[:recorded_date].+(Arel::Nodes.build_quoted('3600 seconds'))
+        Arel::Nodes::SqlLiteral.new('audio_recordings.recorded_date + CAST(audio_recordings.duration_seconds || \' seconds\' as interval)')
```
atruskie, line 196: seems out of place here

app/controllers/audio_event_reports_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+  def respond_report(content, opts = {})
+    content_type = 'application/json'
+    built_response = Settings.api_response.build(:ok, content)
+
+    if request.head?
+      head :ok, { content_length: built_response.to_json.bytesize, content_type: }
+    else
+      render json: built_response, status: :ok, content_type:, layout: false
+    end
+  end
```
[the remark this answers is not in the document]
andrew-1234: fixed (I think?)  changed this to just:
 `render json: result, status: :ok`

config/routes.rb
```diff
@@ -781,6 +781,8 @@ def matches?(request)
       concerns: [:filterable]
   end
 
+  post '/audio_event_reports' => 'audio_event_reports#filter', defaults: { format: 'json' }
```
atruskie, line 784:
```suggestion
  post '/reports/audio_event_summary' => 'reports#audio_event_summary', defaults: { format: 'json' }
```

spec/factories/audio_recording_factory.rb
```diff
@@ -105,7 +105,19 @@
       end
     end
 
+    trait :with_date_range do
+      transient do
+        start_date { '2025-01-01 T00:00:00Z' }
```
atruskie, line 110: if you just make this a date value you don't have to parse it on each sequence generation step

spec/lib/modules/report/.time_series_spec.rb
atruskie, line 1: ignoring

app/modules/report/audio_event_report.rb
```diff
@@ -0,0 +1,259 @@
...
+      time_series_options = TimeSeries::Options.call(@parameters, base_table: BASE_TABLE)
+      composition_options = {
+        bucketed_time_series: Section::Accumulation::TABLE_BUCKETED_TIME_SERIES,
+        base_table: BASE_TABLE,
+        base_verification: Section::EventSummary::TABLE_VERIFICATION_BASE
+      }
+      recording_coverage_options = Report::Section::Coverage.options(time_series_options) { |opt|
+        opt[:source] = base_table
+        opt[:lower_field] = base_table[:recorded_date]
+        opt[:upper_field] = Report::ArelHelpers.arel_recorded_end_date(base_table) # using AudioRecording.arel_recorded_end_date => missing FROM-clause entry for table "audio_recordings"
```
atruskie, line 27: make the function on AudioRecording accept an optional arel_table argument. If not specified just defaults to `self.class.arel_table`

app/modules/report/audio_event_report.rb
```diff
@@ -0,0 +1,259 @@
...
+        opt[:analysis_result] = false
```
atruskie, line 28: leaky

atruskie (commented):
Generally, unless you have calendar specific dates, use `Time` instead of `DateTime`

I've only had time for a partial review. I'll pick this up tomorrow.

app/modules/report/section/coverage.rb
```diff
@@ -0,0 +1,364 @@
...
+        gap_size: TABLE_GAP_SIZE,
+        sort_with_lag: TABLE_SORT_WITH_LAG,
+        grouped: TABLE_GROUPED,
+        coverage_intervals: TABLE_COVERAGE_INTERVALS,
+        coverage_events: TABLE_COVERAGE_EVENTS,
+        coverage_events_sorted_sum: TABLE_COVERAGE_EVENTS_SORTED_SUM,
+        covered_intervals: TABLE_COVERED_INTERVALS,
+        covered_durations: TABLE_COVERED_DURATIONS,
+        final_coverage: TABLE_FINAL_COVERAGE
+      }.freeze
```
atruskie, line 52: i've seen this structure a few times, seems a little redundant? what is it used for?

if you need a list of tables you could register them in your `step` method

app/modules/report/section/step.rb
atruskie, line 1: i think the placement of this class is confusing. It should be next to section , rather than the sections

app/modules/report/table_expression.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+module Report
+  module TableExpression end
+end
```
atruskie, line 5: redundant?

app/modules/report/table_expression/collection.rb
```diff
@@ -0,0 +1,116 @@
...
+module Report
+  # I don't really like having this namespace maybe the classes can be moved
+  module TableExpression
+    # Facilitates the management of CTEs (Report::TableExpression::Datum)
+    # Provides methods to retrieve CTEs in topological order, based on their
+    # dependencies.
+    class Collection
+      def initialize(collection = {})
+        validate collection
+        @entries = collection.is_a?(Hash) ? collection.dup : collection.entries.dup
```
atruskie, line 12: why not enforce one type of collection?

atruskie: why duping? expensive if not needed

app/modules/report/table_expression/collection.rb
```diff
@@ -0,0 +1,116 @@
...
+        required_keys = resolve_dependency_graph(key)
+        get(required_keys)
+      end
+
+      # Return a new collection containing only the specified entries.
+      #
+      # @param keys [Array<Symbol>] The keys to subset the collection with
+      # @return [Report::Expression::Collection] New collection subset
+      def get(keys)
+        input = keys.is_a?(Array) ? keys.map(&:to_sym) : [keys.to_sym]
```
atruskie, line 51: you can use the `Array(keys)` trick here

app/modules/report/table_expression/datum.rb
```diff
@@ -0,0 +1,31 @@
...
+
+      # @return [Arel::Table] The Arel table for the CTE
+      attr_reader :table
+
+      # @return [Arel::Nodes::As] The CTE node
+      attr_reader :cte
+
+      # @return [Array<Symbol>] The names of CTE dependencies
+      attr_reader :depends_on
+    end
```
atruskie, line 29: `Data.define` might be a good candidate for this class - if you keep it as is.

Though it could expand to be a lot more. Imagine if all your `step`s were just classes that inherited from this one?

`Datum` is not a great name for this structure. Maybe `CteNode`? Because nodes in a graph...

spec/lib/modules/report/collection_spec.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+describe 'TableExpression::Collection' do
+  # the correct order is A -> B -> C -> E -> D
+  test_data = [['table_a'], ['table_b'], ['table_c'], ['table_d'], ['table_e']]
+  dependencies = [[], [:table_a], [:table_b], [:table_a, :table_e], [:table_c]]
```
atruskie, line 6: why not have a set of your `Datum` classes here?

spec/lib/modules/report/collection_spec.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+describe 'TableExpression::Collection' do
+  # the correct order is A -> B -> C -> E -> D
```
atruskie, line 4: From a documentation perspective, maybe show the DAG:

A -> B -> C -> E -> D
\\---------------------^

spec/lib/modules/report/collection_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+  end
+
+  describe '#select' do
+    it 'returns a valid select statement for the specified key, with ordered CTEs' do
+      selected = collection.select(:table_c)
+      selected.to_sql
+      expected_sql = <<~SQL.squish
+        WITH "table_a" AS (SELECT), "table_b" AS (SELECT), "table_c" AS (SELECT) SELECT * FROM "table_c"
+      SQL
+      expect(selected.to_sql).to eq(expected_sql)
```
atruskie, line 35: nice

atruskie (changes requested):
I think i was expecting your ctes/steps to be in different files. 

do have any common ctes between different sections?

app/modules/report/audio_event_report.rb
```diff
@@ -0,0 +1,259 @@
...
+    end
+
+    def self.calculate_ratio(count, total)
+      return 0.0 if total.zero?
+
+      (count.to_f / total).round(2)
+    end
+
+    def self.transform_tsrange(row)
+      matches = row['range'].match(/\["([^"]+)","([^"]+)"\)/)
```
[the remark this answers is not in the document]
andrew-1234: using this worked thanks. the output now looks like:
`{:range=>"2025-01-03 00:00:00 UTC...2025-01-03 16:40:00 UTC",  .. }`

instead of an array before `['2025-01-02T00:00:00.000+00:00', '2025-01-03T00:00:00.000+00:00']`
is that good?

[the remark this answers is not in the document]
atruskie: Yeah so, I think what's happening is we get a Ruby range, and the default JSON serializer is to_s/inspecting the value.

It's not great for two reasons: 1) the `...` synax is a distinctly Ruby idiom (AFAIK?) and 2) The date strings are a different format.

I'd either register a JSON serializer for the Range object to emit the tuple variant or do a post query transform on range objects. 

Or we could not use the postgres serializer.


---

Aside: using a two-values tuple for spans/intervals is in no way standard, but it is something I've seen used commonly in D3. I'm opting for that pattern because it is simple and space efficient in JSON, but it's arguably not self-describing in any way.

atruskie:
> > I think i was expecting your ctes/steps to be in different files.
> > do have any common ctes between different sections?
> 
> yea there are at least a couple of CTEs that are defined in one section and required by a different section. would make sense to have these in their own files? in which case, should all ctes be in their own files? unless they are tightly coupled, then they could share a file?
> 
> That way a section wouldn't need to include all of the specific steps/CTE logic (like they do now), they could just define the main select statement needed to get a certain result (e.g. `project.composition.array_agg.as('composition')`, and the collection of CTEs required to execute it.
> 
> And then a concrete report implementation is just a template of which sections to include.
> 
> CTEs -> section = Collection of CTEs -> report = collection of sections

all in their own files? it's a bit overkill but for anything remotely complex it's more maintainable. I would, I think.

the rest sounds good!

app/modules/report/arel_helpers.rb
```diff
@@ -0,0 +1,22 @@
...
+    # Aggregate distinct values for a field.
+    # @param table [Arel::Table] The table to aggregate from
+    # @param field [Symbol] The field to aggregate
+    # @return [Arel::Nodes::NamedFunction] An aggregation node
+    def aggregate_distinct(table, field)
+      Arel::Nodes::NamedFunction.new(
+        'ARRAY_AGG',
+        [Arel::Nodes::SqlLiteral.new("DISTINCT #{table[field].name}")]
+      )
+    end
```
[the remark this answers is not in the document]
andrew-1234: nice, added the distinct helper on attribute and now can compose like:
`base_table[:site_id].distinct.array_agg.as('site_ids')`

app/modules/report/section/composition.rb
```diff
@@ -0,0 +1,137 @@
...
+          )
+          .order(Arel.sql('distinct_tags.tag_id'), bucketed_time_series[:bucket_number])
+      }
+
+      # aggregates the composition series as JSON for reporting
+      def project(_collection = nil, _options = nil)
+        composition_series_aliased = TABLE_COMPOSITION_SERIES.as('c')
+        Arel::SelectManager.new
+          .project(composition_series_aliased.right.json_agg)
+          .from(composition_series_aliased)
```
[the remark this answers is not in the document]
andrew-1234: nah, the 'composition_series_aliased' is an arel::nodes::as so you get undefined method `project` 
there's probably a better way than this weird pattern but i'm not sure how

atruskie:
We went with a different architecture after doing real world performance benchmarks on production data. Still this work was an invaluable first iteration, thank you.

PR #772, opened 2025-06-10
Remove invalid sites from Harvest mappings

### LINKED ISSUE #673: If a site is deleted, and there are mappings in a harvest, the harvest is no longer valid and cannot be used
Steps:

1. Create a project, regions, and site
2. Create an upload in the project
3. Delete the site
4. Now observe the upload cannot be transitioned (e.g. from uploading to scanning) because it is invalid because it contains an invalid mapping.
image

Possible solutions:
- delete mappings from harvests when a site is deleted?
- have the model automatically normalize mappings on transition to remove deleted mappings (the validation can remain the same)?
- Have the client expose a way to edit the mappings in the upload screen? Why get the user to fix this though, we already should have

Related to #672

andrew-1234:
# Remove invalid sites from Harvest mappings

If a site is deleted, and there are mappings in a harvest, the harvest is no longer valid and cannot be used. The purpose of this PR is to implement a callback that removes any invalid site IDs from Harvest mappings, before validation occurs. However, user changes to mappings (mappings_changed?) should still fail with invalid site ids. 

Fixes #673

## Changes

- adds a before_validation callback to clean_up_mappings, unless mappings have changed
- adds relevant request specs
- unrelated factory cascade fix in audio_recordings_spec

## Problems

None identified

## Final Checklist

### DIFF
--- a/app/models/harvest.rb
+++ b/app/models/harvest.rb
@@ -80,7 +80,7 @@ class Harvest < ApplicationRecord
     # we want user changes to mappings to trigger validation
     return if mappings_changed? || mappings.blank?
 
-    mappings_site_ids = mappings.map(&:site_id).compact
+    mappings_site_ids = mappings.map(&:site_id).compact.uniq
     invalid_sites_ids = mappings_site_ids.excluding(
       Site.where(id: mappings_site_ids).pluck(:id)
     ).to_set
--- a/spec/requests/harvest/harvest_spec.rb
+++ b/spec/requests/harvest/harvest_spec.rb
@@ -304,8 +304,9 @@ describe 'Harvesting files' do
         site1.delete
       end
 
+      # before mappings were cleaned up we would fail validation if a site was missing
       it 'cleans up mappings referencing deleted sites when saved' do
-        harvest.save
+        harvest.save!
 
         expect(harvest.mappings.length).to eq 1
         expect(harvest.mappings).to match(a_collection_containing_exactly(


app/models/harvest.rb
```diff
@@ -75,6 +76,18 @@ def validate_uploads_enabled
     errors.add(:project, 'A harvest cannot be created unless its parent project has enabled audio upload')
   end
 
+  def clean_up_mappings
+    # we want user changes to mappings to trigger validation
+    return if mappings_changed? || mappings.blank?
+
+    mappings_site_ids = mappings.map(&:site_id).compact
```
atruskie, line 83: Should we .uniq this list as well?

spec/requests/audio_recordings_spec.rb
atruskie, line 1: Great spot

spec/requests/harvest/harvest_spec.rb
```diff
@@ -293,6 +293,61 @@
...
+
+      before do
+        create_harvest(streaming: false)
+        expect(harvest.mappings.length).to eq 2
+
+        site1.delete
+      end
+
+      it 'cleans up mappings referencing deleted sites when saved' do
+        harvest.save
```
atruskie, line 308: Maybe .save!

Also add a comment like

    # before mappings were cleaned up we would fail validation if a site was missing

atruskie (approved):
Great!

PR #922, opened 2026-03-11
Client-routes

### LINKED ISSUE #654: Verify no links sent to users are aimed at rails sites
- user confirmation email sends user to old site
- permissions email sends users to old site
- analysis job complete email?
- check others

andrew-1234:
# Add Client Routes 

The purpose is to fix links sent to users that point to the old rails site, by directing them to the client instead. Closes #654 

## Changes

Add a module for generating client URLs from default settings (client host, port, protocol)  
Update user confirmation and analysis jobs mail URLs  

Unrelated, add developer dependency gem `ruby-lsp-rspec`  

## Problems

No problems identified.

## Visual Changes

No visual changes.

## Final Checklist

### DIFF
--- a/config/settings/default.yml
+++ b/config/settings/default.yml
@@ -208,7 +208,7 @@ upload_service:
   sftp_port: 2022
 # Configuration for internal microservices
 internal:
-  # array of IPs that we'll accept connect[ions from
+  # array of IPs that we'll accept connections from
   # can be any format that IPAddr.new supports
   allow_list:
     - 172.0.0.0/8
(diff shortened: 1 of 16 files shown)


config/settings/default.yml
```diff
@@ -324,3 +324,9 @@ google_maps:
   api_key: your google maps api key
 permissions:
   any_user_can_create_projects: true
+# Used by Client::Routes to generate external URLs used in emails and other places
+# where we want to link directly to the client.
+client:
+  host: ecosounds.org
```
atruskie: So the default here shouldn't be a production brand. (Ecosounds or a2o). This is generic software until it is branded.

The angular localhost is some port (I forget which) but if we actually wanted to make this work in a development environment then the default settings should be something like https://localhost:4200

PR #975, opened 2026-04-24
feat: add reports/event_summaries endpoint

andrew-1234:
# feat: add reports/event_summaries endpoint

Add a new `POST /reports/event_summaries` endpoint that returns a structured
report of audio event summaries grouped by tag and provenance. Each group
includes score summary statistics (mean, stddev, min, max), a 50-bin score
histogram, plus underflow/overflow counts. Can power visualisations such as score
histogram plots per grouping.

# Changes

- Add `event_summaries` route and action to `ReportsController`
- Add `Api::Reporting::EventSummaries` - report query template that produces event counts, score statistics, and a score histogram per tag/provenance grouping using a series of CTEs
- Add `Api::Reporting::ResultFormatters` - mixin providing `format_score_histogram`, a result-row transformer that reshapes the summary JSON from the query into a structured `score_histogram` key alongside top-level score stats
- Add `execute_report` block argument support in `Api::Reporting` so result rows can be mapped through a formatter
- Add API specs and request specs for `event_summaries`
- Update report permissions spec with custom action for `event_summaries`
- Update routing spec for `event_summaries`

## Problems

No problems identified.

## Visual Changes

No visual changes.

## Final Checklist

### DIFF
--- a/app/controllers/reports_controller.rb
+++ b/app/controllers/reports_controller.rb
@@ -1,8 +1,44 @@
 # frozen_string_literal: true
 
+# Provides reporting endpoints for audio event analysis
 class ReportsController < ApplicationController
   include Api::ControllerHelper
   include Api::Reporting
+  include ResultFormatters
+
+  # POST /reports/event_summaries
+  # Returns a structured report of event summaries per tag and provenance groupings.
+  # Accepts a filter object where:
+  #  the `filter` is applied to audio events
+  #  the `paging`, `sort` and `projection` options are invalid
+  def event_summaries
+    do_authorize_class(:filter, AudioEvent)
+
+    base_query = Access::ByPermissionTable.audio_events(current_user, level: Access::Permission::READER)
+
+    event_summaries_template = EventSummaries.new
+
+    projections = {
+      summary: event_summaries_template.event_summary
+    }
+
+    results, opts = execute_report(
+      base_query:,
+      template: event_summaries_template,
+      projections:
+    )
+
+    score_keys = [:score_mean, :score_stddev, :score_minimum, :score_maximum, :score_histogram]
+
+    results.map do |row|
+      row => { summary:, **rest }
+
+      histogram = extract_histogram(summary, histogram_key: :score_histogram)
+      rest.merge(ensure_columns(score_keys, histogram))
+    end => results
+
+    respond_report(results, opts)
+  end
 
   # POST /reports/tag_diel_activity
   # Returns a structured report of tag frequencies over diel time buckets.
--- a/app/modules/api/reporting/event_summaries.rb
+++ b/app/modules/api/reporting/event_summaries.rb
@@ -1,0 +1,251 @@
+# frozen_string_literal: true
+
+module Api
+  module Reporting
+    # Report template with the required CTEs and joins to produce a report of
+    # event summaries per groupings (tag, provenance). Includes event score
+    # summary statistics, and a score histogram with event counts per bucket,
+    # calculated as the relative position of the score between the histogram's
+    # minimum and maximum score.
+    #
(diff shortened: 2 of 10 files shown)


spec/requests/reports/event_summaries_spec.rb
```diff
@@ -0,0 +1,152 @@
...
+    # create an audio_event that writer_user has no access to, to prove it is not included in the counts
+    create(:audio_event_with_tags)
+  end
+
+  it 'returns the correct summary data' do
+    post '/reports/event_summaries', params: { filter: {} }, **api_headers(writer_token)
+    expect_success
+
+    expect(api_data).to match(array_including(expected_data))
+  end
```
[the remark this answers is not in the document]
andrew-1234: array_including unwraps passed arrays

app/controllers/reports_controller.rb
```diff
@@ -1,8 +1,36 @@
...
+
+    projections = {
+      summary: event_summaries_template.event_summary
+    }
+
+    results, opts = execute_report(
+      base_query:,
+      template: event_summaries_template,
+      projections:,
+      &format_score_histogram
```
atruskie: Ok another way to do this is (without the include):: `&ResultFormatters.format_score_histogram`

app/modules/api/reporting/event_summaries.rb
```diff
@@ -0,0 +1,245 @@
...
+          .when(score_lt_histogram_min?).then(UNDERFLOW_BIN_INDEX)
+          .when(score_gt_histogram_max?).then(NUMBER_OF_BINS)
+          .else(Arel.null)
+      end
+
+      def score_gteq_histogram_min? = EVENTS[:score].gteq(STATS[HISTOGRAM_MINIMUM])
+      def score_lt_histogram_max? = EVENTS[:score].lt(STATS[HISTOGRAM_MAXIMUM])
+      def score_eq_histogram_max? = EVENTS[:score].eq(STATS[HISTOGRAM_MAXIMUM])
+      def score_lt_histogram_min? = EVENTS[:score].lt(STATS[HISTOGRAM_MINIMUM])
+      def score_gt_histogram_max? = EVENTS[:score].gt(STATS[HISTOGRAM_MAXIMUM])
```
atruskie: nit: while I appreciate that these are SQL predicates, they're not ruby predicates - as in the methods do not return TrueClass or FalseClass. Probably should not suffix them with a `?`

spec/api/reports/event_summaries_spec.rb
```diff
@@ -0,0 +1,155 @@
...
+describe 'reports', type: :request do
+  create_entire_hierarchy
+
+  sends_json_and_expects_json
+  with_authorization
+
+  let(:skip_automatic_description) { true }
+
+  def self.response_body_schema
+    Api::Schema.standard_array_response(
```
atruskie, line 27: ok I know this isn't a new pattern, so no need to act, 

but I don't like the pattern where api schemas are defined in the test files.

The schemas should be defined as part of the application code - so that they're available to application code to use for other purposes - and not be part of the test spec.

That's why I have a `.schema` method on all the models. 

This is here because there is no natural spot to put the schemas for these APIs... 

Tl;DR: this is an anti-pattern, but we'll fix it later

atruskie (changes requested):
We talked offline about simplifying some of the reporting transforms.

It's also a choice to return json sub-objects rather than just columns, with different tradefoffs.  It's a bit of an anit-pattern, but performance is identical and @andrew-1234 explains that it vastly simplifies code-gen for an edge case (only one case-when statement needed for nil scores, rather than one set per column). We should add a comment to that effect in the module that creates the json.

Otherwise, great work.

atruskie (approved):

PR #990, opened 2026-05-12
feat: add reports/recording_coverage, reports/analysis_coverage endpoints; upgrade db image to postgres:18.3

andrew-1234:
# feat: add reports/recording_coverage, reports/analysis_coverage endpoints; upgrade db image to postgres:18.3

The purpose is to add coverage reporting endpoints for recordings and recording analysis results.  

The db version is also updated to 18.3, which allows using the range_agg function (simplifies the coverage query), and CTE heavy queries run ~4x faster.

## Changes

- Add `POST /reports/recording_coverage` and `POST /reports/analysis_coverage` to `ReportsController`, with routing, permissions, request specs, routing specs, and api specs.
- Add `Api::Reporting::Coverage` to calculate contiguous coverage spans per configurable partition 
- Update `Api::Reporting.execute_report` to accept the report model explicitly so coverage reports can reuse the existing reporting pipeline with `AudioRecording` filters and query casting.

- Add custom Arel helpers and SQL visitor support for ordered `array_agg`, `range_agg`, range upper bounds, and `tsmultirange_total_seconds` so the coverage queries can be composed cleanly in Ruby.
- Add the `contiguous_range_number` PostgreSQL aggregate migration to assign island numbers to contiguous `tsrange` groups.
- Add the `tsmultirange_total_seconds` PostgreSQL function migration to sum covered seconds in multiranges.

- Upgrade the database container to PostgreSQL 18.3.
- Update the Postgres volume mount location for the newer image layout and install the PostgreSQL 18 client in the web container.

## Problems

- db major version jump?

## Visual Changes

No visual changes.

## Final Checklist

### DIFF
--- a/app/controllers/reports_controller.rb
+++ b/app/controllers/reports_controller.rb
@@ -6,6 +6,60 @@ class ReportsController < ApplicationController
   include Api::Reporting
   include ResultFormatters
 
+  # POST /reports/recording_coverage
+  # Returns a structured report of recording coverage
+  # Accepts a filter object where:
+  #  the `filter` is applied to audio recordings
+  #  the `paging`, `sort` and `projection` options are invalid
+  def recording_coverage
+    do_authorize_class(:filter, AudioRecording)
+
+    base_query = Access::ByPermissionTable.audio_recordings(current_user, level: Access::Permission::READER)
+
+    partition_columns = [AudioRecording.arel_table[:site_id]]
+
+    recording_coverage_template = Coverage.new(partition_columns: partition_columns, options: coverage_report_options)
+
+    results, opts = execute_report(
+      base_query:,
+      model: AudioRecording,
+      template: recording_coverage_template,
+      projections: { density: Coverage.coverage_density }
+    )
+
+    respond_report(results, opts)
+  end
+
+  # POST /reports/analysis_coverage
+  # Returns a structured report of recording analysis coverage
+  # Accepts a filter object where:
+  #  the `filter` is applied to audio recordings
+  #  the `paging`, `sort` and `projection` options are invalid
+  def analysis_coverage
+    do_authorize_class(:filter, AudioRecording)
+
+    base_query = Access::ByPermissionTable.audio_recordings(current_user, level: Access::Permission::READER)
+
+    partition_columns = [AudioRecording.arel_table[:site_id], AnalysisJobsItem.arel_table[:result]]
+    joins = AudioRecording.arel_table
+      .join(AnalysisJobsItem.arel_table)
+      .on(AnalysisJobsItem.arel_table[:audio_recording_id].eq(AudioRecording.arel_table[:id])
+      .and(AnalysisJobsItem.arel_table[:result].is_not_null))
+      .join_sources
+
+    analysis_coverage_template = Coverage.new(partition_columns: partition_columns, joins: joins,
+      options: coverage_report_options)
+
+    results, opts = execute_report(
+      base_query:,
+      model: AudioRecording,
+      template: analysis_coverage_template,
+      projections: { density: Coverage.coverage_density }
+    )
+
+    respond_report(results, opts)
+  end
+
   # POST /reports/event_summaries
   # Returns a structured report of event summaries per tag and provenance groupings.
   # Accepts a filter object where:
@@ -24,6 +78,7 @@ class ReportsController < ApplicationController
 
     results, opts = execute_report(
       base_query:,
+      model: AudioEvent,
       template: event_summaries_template,
       projections:
     )
@@ -58,6 +113,7 @@ class ReportsController < ApplicationController
 
     results, opts = execute_report(
       base_query:,
+      model: AudioEvent,
       template: TagDielActivity.new(report_options),
       projections:
     )
@@ -83,6 +139,7 @@ class ReportsController < ApplicationController
 
     results, opts = execute_report(
       base_query:,
+      model: AudioEvent,
       template: TagFrequency.new(report_options),
       projections:
     )
@@ -109,6 +166,7 @@ class ReportsController < ApplicationController
 
     results, opts = execute_report(
       base_query:,
+      model: AudioEvent,
       template: TagAccumulation.new(report_options),
       projections:
     )
@@ -117,6 +175,12 @@ class ReportsController < ApplicationController
   end
 
   private
+
+  def coverage_report_options
+    options = api_options_params
+    options.require(:bucket_count)
+    options.permit(:bucket_count).to_h
+  end
 
   # @return [Hash]
   def report_options
--- a/app/modules/api/reporting/coverage.rb
+++ b/app/modules/api/reporting/coverage.rb
@@ -1,0 +1,140 @@
+# frozen_string_literal: true
+
+module Api
+  module Reporting
+    # Coverage periods are calculated as the contiguous ranges formed by a sequence
+    # of overlapping or neighbouring recordings (within a gap threshold), per
+    # customisable partitions (e.g. by site).
+    #
+    # Implements #call(query) for use as a template in execute_report.
+    class Coverage
+      include CteHelper
+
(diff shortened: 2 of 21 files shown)


app/modules/api/reporting/coverage.rb
```diff
@@ -0,0 +1,126 @@
...
+      end
+
+      # Dynamically calculate a gap threshold as 1/1920th of the total span of all recordings in the query
+      def threshold_cte
+        # ! TODO: Division when arel-extensions is removed. See https://github.com/QutEcoacoustics/baw-server/issues/966
+        span = Arel::Nodes::Subtraction.new(
+          RECORDINGS[:recording_range].upper.maximum,
+          RECORDINGS[:recording_range].lower.minimum
+        )
+        RECORDINGS.project(Arel.seconds(Arel::Nodes::Division.new(span.extract('epoch'), 1920)).as('val'))
```
atruskie: okay the 1920 should not be hard coded here. we should be pulling the option out of params. It should be named `bucket_count` or something like that.

app/modules/api/reporting/coverage.rb
```diff
@@ -0,0 +1,126 @@
...
+            recording_range_arel.as('recording_range'),
+            *@partition_columns
+          ).arel
+      end
+
+      def recording_range_arel
+        Arel.tsrange(AudioRecording.arel_table[:recorded_date], AudioRecording.arel_recorded_end_date)
+      end
+
+      # Dynamically calculate a gap threshold as 1/1920th of the total span of all recordings in the query
```
atruskie:
```suggestion
      # Dynamically calculate a gap threshold as 1/1920th of the total span of all recordings in the query
      # We don't want to render spans that are so short that they would occupy less than 1px on the screen.
      # Thus, this option allows us to emit the least number of spans possible, that will actually be renderable.
```

app/controllers/reports_controller.rb
```diff
@@ -6,6 +6,59 @@ class ReportsController < ApplicationController
...
+
+  # POST /reports/analysis_coverage
+  # Returns a structured report of recording analysis coverage
+  # Accepts a filter object where:
+  #  the `filter` is applied to audio recordings
+  #  the `paging`, `sort` and `projection` options are invalid
+  def analysis_coverage
+    do_authorize_class(:filter, AudioRecording)
+
+    base_query = Access::ByPermissionTable.audio_recordings(current_user, level: Access::Permission::READER)
```
atruskie, line 41: missing bucket count option and `report_options` equivalent helper

app/modules/api/schema/helpers.rb
```diff
@@ -186,6 +186,56 @@ def report_options(diel: false)
...
+      def recording_coverage(include_result: false)
+        properties = {
+          site_id: id,
+          coverage: {
+            type: 'array',
+            items: { type: 'string', format: 'date-time' },
+            minItems: 2,
+            maxItems: 2,
+            additionalItems: false,
+            description: 'The start and end of the contiguous coverage span'
```
atruskie: comment on open/closed bounds?
```suggestion
            description: 'The inclusive start and exclusive end of the contiguous coverage span'
```

app/modules/api/schema/helpers.rb
```diff
@@ -186,6 +186,56 @@ def report_options(diel: false)
...
+          density: {
+            type: 'number',
+            description: 'The ratio of covered seconds to the total duration of the coverage span',
+            minimum: 0.0,
+            maximum: 1.0
+          },
+          gap_threshold: {
+            type: 'number',
+            description: 'The maximum number of seconds between neighbouring recordings before a new coverage span begins' \
+                         ' Calculated dynamically as 1/1920th of the total span of all recordings in the query'
```
atruskie: should not be fixed 1920

app/modules/baw/arel/visitors/to_sql_extensions.rb
```diff
@@ -10,6 +10,17 @@ module Visitors
       # An extension to https://github.com/rails/rails/blob/main/activerecord/lib/arel/visitors/to_sql.rb.
       # Only valid for postgesql.
       module ToSqlExtensions
+        def visit_Baw_Arel_Nodes_ArrayAgg(o, collector)
+          return visit_Arel_Nodes_NamedFunction(o, collector) unless o.order_expressions.present?
```
atruskie: seems like a weird diversion. i'd just return early after the expressions are visited if order_expressions.blank?

app/modules/baw/arel.rb
```diff
@@ -71,6 +78,11 @@ def unqualified
       def seconds
         Nodes::MakeInterval.new(seconds: self)
       end
+
+      # Returns the upper bound of a range attribute; Arel already provides `lower`.
+      def upper
+        ::Arel::Nodes::NamedFunction.new('upper', [self])
+      end
```
atruskie, line 85: really? arel has one but not the other???

db/migrate/20260507052638_add_contiguous_range_number_function.rb
```diff
@@ -0,0 +1,94 @@
...
+# interval threshold. Whenever a gap exceeds the threshold the sequence number
+# is incremented, effectively labelling each distinct contiguous block.
+#
+# Objects created (in dependency order):
+#   1. Type      contiguous_range_state        – accumulator state
+#   2. Function  contiguous_range_transition   – aggregate step function
+#   3. Function  contiguous_range_final        – aggregate final function
+#   4. Aggregate contiguous_range_number       – the public aggregate
+class AddContiguousRangeNumberFunction < ActiveRecord::Migration[8.0]
+  def up
```
atruskie, line 11: This is mostly fine but it would be good if we added these comments as comments in the schema.

See example here: https://github.com/QutEcoacoustics/baw-server/wiki/Archive:-deterministic-location-jitter-in-SQL

atruskie: To drive this home the comments you do have in the migration won't survive the schema dump and won't end up in the database

db/migrate/20260507061809_add_tsmultirange_total_seconds_function.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+# Adds a PostgreSQL function `tsmultirange_total_seconds` that accepts a
+# `tsmultirange` and returns the sum of each constituent `tsrange`'s duration in
+# seconds as a double precision value.
+#
+# Example use case is to input the tsmultirange returned by 'range_agg', to get
+# the total non-overlapping seconds covered by the tsmultirange. This is useful
+# because it saves having to handle unnesting the tsmultirange and simplifies the query.
+class AddTsmultirangeTotalSecondsFunction < ActiveRecord::Migration[8.0]
```
atruskie, line 4: Same as previous, emit actual schema comments

atruskie (changes requested):
Awesome

app/modules/baw/arel.rb
andrew-1234: haha yea, and I think just because they use it for converting text to lower

atruskie: `https://github.com/rails/rails/pull/45651` 🥲

app/modules/api/schema/helpers.rb
```diff
@@ -206,8 +220,8 @@ def recording_coverage(include_result: false)
           },
           gap_threshold: {
             type: 'number',
-            description: 'The maximum number of seconds between neighbouring recordings before a new coverage span begins' \
-                         ' Calculated dynamically as 1/1920th of the total span of all recordings in the query'
+            description: 'The maximum number of seconds between neighbouring recordings before a new coverage span begins ' \
+                         'Calculated dynamically as 1/1920th of the total span of all recordings in the query'
```
atruskie: 1920 reference

atruskie (approved):

PR #1005, opened 2026-06-19
feat: implement async email delivery mailers

### LINKED ISSUE #1004: Make all email async
We have a weird situation where our vendor is refusing to fix a firewall rule that is blocking outbound email.

However, some of our servers are not blocked.

If we move our mail delivery to ActiveJob jobs, we can choose which servers deliver mail, and bypass the weird firewall rule.

We've been meaning to do this anyway (to make web requests responsive and non-blocking on external IO), but not it is a priority because of the firewall malarky.

atruskie:
Fixes #1004.

Makes all email run in background jobs.

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -147,7 +147,7 @@ ORDER BY project_count ASC, s.name ASC")
     respond_to do |format|
       if valid_request
         ProjectMailer.project_access_request(current_user, access_request_params[:projects],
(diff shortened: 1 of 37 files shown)


atruskie:
@copilot look at the failing CI logs and fix the errors

atruskie:
@copilot So we skipped emails sent in background workers because they're already in background jobs. But I realised the point of this is to route all mail through one worker which has firewall access to send mail. So we still need to async send mail for those cases as well

Issue #1009, opened 2026-06-30
Update the tags model to support new type_of_tag values


andrew-1234:
With camtrap data package exports, observationType is a required field, and has allowed values 'animal', 'human', 'vehicle', 'blank', 'unknown', 'unclassified'.  type_of_tag: common_name and species_name map to oberservationType 'animal' easily. But the others require pattern matching against a tag's text. E.g. if tag.text == 'car', then it's a 'vehicle'. (For camtrap-dp exports, observations (taggings) that we can't match to an observationType enum value are 'unclassified'.)

Instead of maintaining a mapping from known patterns to their type, we can add new types and update existing tags to use the new types where it makes sense.

E.g. for 'vehicles', add a new value 'vehicle' to type_of_tag enum, and  transform existing tags that match a vehicle pattern from having type_of_tag: general to vehicle.

vehicle patterns
            /airplane/i,
            /boat/i,
            /car/i,
            /engine/i,
            /helicopter/i,
            /motor/i,
            /train/i,
            /train passing/i,
            /train horn/i,
            /machine generated/i,
            /vehicle/i

human patterns
            /baby/i,
            /human voice/i
          
unknown patterns
            /unknown/i,
            /unknown_\d+/i,                   (unknown_1 through to 12 exist in database)
            /unsure/i

Q: Some existing tags have text that could be a vehicle but it's ambiguous, like motor or engine - do we map to vehicle, leave as general, or map to our own new type?

There are also species and common name tags that are marked as general, but that is part of 
https://github.com/QutEcoacoustics/baw-server/issues/991

## Tasks 

- [ ] update https://github.com/QutEcoacoustics/baw-server/wiki/Tag-classification-rules to reflect any changes

atruskie:
Unknown and unsure can map too.

The other option is to add an extra column to the database. It might break fewer things.

I'm not sure the concepts are sufficiently different however.

PR #1011, opened 2026-07-03
feat: add camtrap-dp export module 

### LINKED ISSUE #984: `BawWorkers::Export::CamtrapDp` exporter module.
A class under `lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/`
that, given a filter query + options, produces a Camtrap‑DP (Bioacoustics) zip.

The vendored profile + table schemas live under
`lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/profile/` and are
copied verbatim from `camera-traps/bioacoustics/camtrap-dp/1.0.2/` at a pinned
commit. A rake task `baw:export:camtrap_dp:update_profile` refreshes them.

### Interface

```ruby
BawWorkers::Export::CamtrapDp::Exporter.new(
  filter:               Hash,   # baw-server filter DSL (same shape as analysis_jobs.filter)
  requested_by:         User,   # for permission scoping
  obfuscate_locations:  true,   # respect sites.obfuscated_*
  license_id:           nil,    # override project.license; vocabulary = ALA's
  include_media:        false   # is intentionally not implemented in v1 (stretch goal)
).call(output: IO) # returns {  manifest:, stats: }
```

### Mapping

| Camtrap-DP (Bioacoustics) resource | Source                                                 |
|------------------------------------|--------------------------------------------------------|
| `datapackage.json`                 | Frictionless profile + project metadata + filter echo  |
| `deployments.csv`                  | One row per `(site, deployment_window)` derived from `audio_recordings.recorded_date` + duration grouped by site and upload id |
| `media.csv`                        | `audio_recordings` (URL via existing media route)      |
| `observations.csv`                 | `audio_events` × `audio_events_tags`, joined with `tags`, `provenances`, `verifications` |
| Bioacoustics extension columns     | `start_time_seconds`, `end_time_seconds`, `low_frequency_hertz`, `high_frequency_hertz`, `score`, `channel`, `classifiedBy` (provenance), verification consensus |

### Behaviour

- Streams CSVs to disk to bound memory (use `CSV.open` + `find_each`).
- Honours permissions: filter is first run through `Access::ByEffectivePermission` for
  `requested_by` before producing rows.
- Validates output against the vendored Bioacoustics profile JSON
  (frictionless) and fails the job on schema errors.
- Deterministic: identical inputs → identical zip bytes (sorted rows, stable
  uuids derived from record ids + package_id).
- **License vocabulary**: we assume baw-server `projects.license` already uses
  (or can be coerced to) the ALA‑accepted vocabulary. The exporter passes the
  string through unchanged. Add a non‑blocking warning to `manifest.warnings`
  when the value doesn't match a known token, but do not transform it.

### Tests

- Schema validation against pinned profile fixture.
- Permission scoping (anonymous / logged‑in / project member).
- Round‑trip example in `spec/lib/baw_workers/export/camtrap_dp_spec.rb`.

andrew-1234:
# feat: add camtrap-dp export module 

The purpose is to export validated Camtrap Data Packages from filtered audio event taggings. Fixes #984 
## Changes

Add a `BawWorkers::Export::CamtrapDp` exporter that writes deployments, media, observations, and `datapackage.json` into a zipped package
Add Camtrap DP descriptor, table, metadata, validator, error, and profile modules backed by dry-struct types
Add pinned acoustic Camtrap DP profile assets and a `baw:camtrap_dp:update` task for refreshing the local validation profile
Add DataPackage/TableSchema patches needed for local profile loading and schema-compatible CSV validation

Add export options for coordinate obfuscation, contributors, project metadata, project license emission, and forced UTC offsets
Add timezone-aware deployment accumulation using site timezones with UTC fallback
Map audio recordings and taggings into Camtrap DP media and interval observation rows
Derive package spatial, temporal, taxonomic, project, and license metadata from exported deployments

Add `AudioRecording#recorded_end_date`, `Permission#broad_access?`, and `Site#public_site?` helpers used by export mapping
Update audio recording overlap logic to reuse `recorded_end_date`

Add specs for exporter validation and package output, observation type mapping, profile update task, and the new model helpers

## Problems

Observation type mapping is temporary and should be revisited with #1009.

## Visual Changes

No visual changes.

## Final Checklist

### DIFF
--- a/app/models/permission.rb
+++ b/app/models/permission.rb
@@ -48,6 +48,12 @@ class Permission < ApplicationRecord
     { id: l, name: Access::Core.get_level_name(l) }
   }
 
+  # Returns true when this permission grants access to anonymous or logged-in users.
+  # @return [Boolean]
+  def public_access?
+    (allow_anonymous || allow_logged_in) && Access::Core.levels.include?(level&.to_sym)
+  end
+
   # association validations
   # these validations are now redundant i think
   #validates_associated :project
--- a/lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/deployment_accumulator.rb
+++ b/lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/deployment_accumulator.rb
@@ -1,0 +1,98 @@
+# frozen_string_literal: true
+
+module BawWorkers
+  module Export
+    module CamtrapDp
+      # Accumulate deployment related metadata for taggings.
+      #
+      # We use Site as a proxy for a deployment, and calculate deployment start and end time based on all audio
+      # recordings at the site, because we don't track deployments in the database.
+      class DeploymentAccumulator
+        # The caller can provide a timezone override. If provided, all timestamps in the export will use this timezone.
+        # You can even provide a custom timezone, e.g.: `ActiveSupport::TimeZone.create('Fixed', 600, TZInfo::Timezone.get('GMT'))`
+        #
+        # @param forced_timezone [ActiveSupport::TimeZone, TZInfo::Timezone, nil] optional timezone
+        def initialize(forced_timezone: nil)
+          @deployments = {}
+
+          if forced_timezone && !(forced_timezone.is_a?(ActiveSupport::TimeZone) || forced_timezone.is_a?(TZInfo::Timezone))
+            raise ArgumentError, "forced_timezone: got #{forced_timezone.class}, expected ActiveSupport::TimeZone or TZInfo::Timezone"
+          end
+
+          @forced_timezone = forced_timezone
+        end
+
+        # Deployment metadata accumulated for a site.
+        #
+        # @!attribute [r] site
+        #   @return [Site]
+        # @!attribute [r] start
+        #   @return [Time] earliest recording start time for the deployment.
+        # @!attribute [r] end
+        #   @return [Time] latest recording end time for the deployment.
+        # @!attribute [r] file_public
+        #   @return [Boolean] whether media files for the deployment are publicly accessible.
+        # @!attribute [r] timezone
+        #   @return [TZInfo::Timezone, ActiveSupport::TimeZone] the timezone to use for timestamps in this deployment
+        Deployment = Data.define(:site, :start, :end, :file_public, :timezone) {
+          # Ensures all times in a deployment are in a uniform Timezone for exporting.
+          #
+          # @param time [Time, ActiveSupport::TimeWithZone, nil] timestamp to export.
+          # @return [Time, ActiveSupport::TimeWithZone, nil] timestamp converted to the deployment's timezone,
+          #   ready for Camtrap-DP ISO 8601 serialization.
+          def ensure_timezone(time)
+            return nil if time.blank?
+
+            time.in_time_zone(timezone)
+          end
+        }
+
+        # Add a deployment for the tagging's site if it doesn't exist. Initialize the deployment's timezone based on the
+        # site's timezone, unless a forced_timezone was provided to the accumulator.
+        #
+        # This isn't a problem now, but it's worth noting: deployments are keyed by site_id, and site is cached the
+        # first time seen. So when using find_each in batches, the same site from a later batch will have a different
+        # object_id than the cached site object.
+        #
+        # @param tagging [Tagging] the tagging for the deployment
+        # @return [Deployment] the new or updated deployment for the tagging's site
+        def add_or_update(tagging)
+          audio_recording = tagging.audio_event.audio_recording
+          @deployments[audio_recording.site_id] ||= build_deployment(audio_recording.site)
+        end
+
+        def values = @deployments.values
+
+        private
+
+        # Build a deployment for a site, calculating the start and end time
+        # based on all audio recordings at the site.
+        # @param site [Site] the site for which to build a deployment
+        # @return [Deployment] the deployment for the site
+        def build_deployment(site)
+          timezone = @forced_timezone || site_timezone(site)
+          recording_bounds = AudioRecording.pick_hash({
+            start: AudioRecording.arel_table[:recorded_date].minimum,
+            end: AudioRecording.arel_recorded_end_date.maximum
+          }, scope: site.audio_recordings)
+
+          Deployment.new(
+            site: site,
+            start: recording_bounds[:start]&.in_time_zone(timezone),
+            end: recording_bounds[:end]&.in_time_zone(timezone),
+            file_public: site.public_site?,
+            timezone: timezone
+          )
+        end
+
+        # @param site [Site] site whose timezone should be used for export timestamps.
+        # @return [TZInfo::Timezone, ActiveSupport::TimeZone] timezone of site or UTC fallback if site has no timezone.
+        def site_timezone(site)
+          return ActiveSupport::TimeZone['UTC'] if site.timezone.blank?
+
+          TimeZoneHelper.tzinfo_class(site.tzinfo_tz)
+        end
+      end
+    end
+  end
+end
--- a/lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/descriptor/contributor.rb
+++ b/lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/descriptor/contributor.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+module BawWorkers
+  module Export
+    module CamtrapDp
+      module Descriptor
+        # implements: camtrap-dp-profile-acoustic.json#/allOf/1/properties/contributors/items
+        class Contributor < Base
+          # name/title of the person or organisation
+          attribute :title, Types::String
+          attribute :role, Types::Role
+
+          attribute? :email, Types::String
+          attribute? :path, Types::Url
+          attribute? :organization, Types::String
+        end
+      end
+    end
+  end
+end
--- a/lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/descriptor/license.rb
+++ b/lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/descriptor/license.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+module BawWorkers
+  module Export
+    module CamtrapDp
+      module Descriptor
+        # implements: camtrap-dp-profile-acoustic.json#/allOf/1/properties/licenses/items
+        class License < Base
+          attribute :name, Types::String
+          attribute? :path, Types::UrlOrPath
+          attribute? :title, Types::String
+          attribute :scope, Types::String.enum('data', 'media')
+        end
+      end
+    end
+  end
+end
--- a/lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/descriptor/package.rb
+++ b/lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/descriptor/package.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+module BawWorkers
+  module Export
+    module CamtrapDp
+      module Descriptor
+        # implements: camtrap-dp-profile-acoustic.json#/allOf/1/properties
+        class Package < Base
+          attribute :profile, Types::Url
+          attribute :resources, Types::Array.of(Resource)
+          attribute :created, Types::UtcTimeSeconds
+          attribute :contributors, Types::Array.of(Contributor)
+          attribute :project, Project
+          attribute :spatial, Types::GeoJSON
+          attribute :temporal, Temporal
+          attribute :taxonomic, Types::Array.of(Taxonomic)
+
+          attribute? :name, Types::String
+          attribute? :id, Types::String
+          attribute? :title, Types::String
+          attribute? :description, Types::String
+          attribute? :version, Types::String
+          attribute? :keywords, Types::Array.of(Types::String)
+          attribute? :image, Types::String
+          attribute? :homepage, Types::String
+          attribute? :sources, Types::Array.of(Source)
+          attribute? :licenses, Types::Array.of(License)
+          attribute? :bibliographicCitation, Types::String
+          attribute? :coordinatePrecision, Types::Float
+          attribute? :relatedIdentifiers, Types::Array.of(RelatedIdentifier)
+          attribute? :references, Types::Array.of(Types::String)
+        end
+      end
+    end
+  end
+end
--- a/lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/descriptor/resource.rb
+++ b/lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/descriptor/resource.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+module BawWorkers
+  module Export
+    module CamtrapDp
+      module Descriptor
+        # implements: camtrap-dp-profile-acoustic.json#/allOf/1/properties/resources/items/oneOf/0
+        class Resource < Base
+          attribute :name, Types::String
+          attribute :path, Types::UrlOrPath
+
+          # The resources in camtrap-dp are tabular-data-resources, so this is a fixed value.
+          attribute :profile, Types::String.default('tabular-data-resource')
+          attribute :format, Types::String.default('csv')
+          attribute :mediatype, Types::String.default('text/csv')
+          attribute :encoding, Types::String.default('utf-8')
+
+          # The raw parsed JSON table schema or url-or-path to the schema
+          attribute :schema, Types::Schema
+        end
+      end
+    end
+  end
+end
--- a/lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/descriptor/source.rb
+++ b/lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/descriptor/source.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+module BawWorkers
+  module Export
+    module CamtrapDp
+      module Descriptor
+        # Represents a source of the package, e.g. a data source. Can be a data management platform from which the
+        # package was derived.
+        # implements: camtrap-dp-profile-acoustic.json#/allOf/1/properties/sources/items
+        class Source < Base
+          attribute? :title, Types::String
+          attribute? :path, Types::UrlOrPath
+          attribute? :email, Types::String
+          attribute? :version, Types::String
+        end
+      end
+    end
+  end
+end
--- a/lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/descriptor/temporal.rb
+++ b/lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/descriptor/temporal.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+module BawWorkers
+  module Export
+    module CamtrapDp
+      module Descriptor
+        # implements: camtrap-dp-profile-acoustic.json#/allOf/1/properties/temporal
+        class Temporal < Base
+          attribute :start, Types::UtcTimeSeconds
+          attribute :end, Types::UtcTimeSeconds
+        end
+      end
+    end
+  end
+end
--- a/lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/exporter.rb
+++ b/lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/exporter.rb
@@ -1,0 +1,248 @@
+# frozen_string_literal: true
+
+require_relative 'errors/camtrap_dp_error'
+module BawWorkers
+  module Export
+    module CamtrapDp
+      # NOTE: ALA exports will be limited to a single project, enforced upstream, not in this module.
+      # NOTE: :should_obfuscate option is respected by this module; the ability will be enforced in the controller
+      #
+      # The exporter orchestrates a Camtrap Data Package export from a given filter of taggings.
+      # It generates a zip file containing the CSV tables and the datapackage.json file describing the package.
+      class Exporter
+        BATCH_SIZE = 1000
+
+        # Options required to build a complete Camtrap Data Package export.
+        #
+        # @!attribute [r] user
+        #   @return [User, nil] optional user requesting the export, to check permissions against
+        #     (e.g. public_latitude, public_longitude).
+        # @!attribute [r] should_obfuscate
+        #   @return [Boolean, nil] optional, override the default obfuscation behaviour and indicate whether
+        #     or not to obfuscate locations.
+        # @!attribute [r] contributors
+        #   @return [Array<Hash>] descriptor-compatible package contributors.
+        # @!attribute [r] project_capture_method
+        #   @return [Array<String>] project-level Camtrap DP captureMethod values (see Types::CaptureMethod).
+        # @!attribute [r] project_sampling_design
+        #   @return [String] project-level Camtrap DP samplingDesign value (see Types::SamplingDesign).
+        # @!attribute [r] package_title
+        #   @return [String] Optional title for this specific package, different from the project title
+        # @!attribute [r] emit_project_license
+        #   @return [Boolean] whether to include project license metadata in the package descriptor.
+        # @!attribute [r] forced_timezone
+        #   @return [ActiveSupport::TimeZone, TZInfo::Timezone, nil] optional timezone to apply to every
+        #     exported recording, event, deployment, and temporal coverage timestamp. When omitted, each
+        #     deployment uses its site's timezone; sites with no timezone or a zero UTC offset export in UTC.
+        RequiredExporterOptions = Data.define(
+          :user,
+          :should_obfuscate,
+          :contributors,
+          :project_capture_method,
+          :project_sampling_design,
+          :package_title,
+          :emit_project_license,
+          :forced_timezone
+        )
+
+        # @param filter [ActiveRecord::Relation<Tagging>] a filter of taggings to export
+        # @param exporter_options [RequiredExporterOptions] options required to build a complete Camtrap Data Package
+        def initialize(filter, exporter_options)
+          validate_filter(filter)
+          validate_exporter_options(exporter_options)
+
+          @filter = filter
+          @options = exporter_options
+        end
+
+        # Generates the export and yields a manifest of metadata about the generated export to the provided block.
+        #
+        # The manifest's `package_path` points to the directory containing the unpacked Camtrap Data Package, while
+        # `zip_path` points to an archive containing the same files. The parent temporary directory is deleted when the
+        # block returns, so the archive should be copied before then.
+        #
+        # The generated files have this layout:
+        #
+        #   tmp/exporter_<timestamp>-<random>/
+        #   ├── dp/
+        #   │   ├── datapackage.json
+        #   │   ├── deployments.csv
+        #   │   ├── media.csv
+        #   │   └── observations.csv
+        #   └── dp.zip
+        #
+        # @yield [PackageManifest] export metadata:
+        #   - `package_path` [Pathname] directory containing the package files
+        #   - `zip_path` [Pathname] path to the zip archive
+        #   - `file_stats` [Hash<Symbol, Hash>] size and modification time for each package file
+        # @return the result of the block is returned to the caller
+        def call(&block)
+          raise ArgumentError, 'block is required' unless block_given?
+
+          # Bullet doesn't see the associations being used below and wrongly raises an UnoptimizedQueryError:
+          # avoid eager loading.
+          bullet_active = defined?(Bullet) && Bullet.enabled?
+          Bullet.enable = false if bullet_active
+
+          Rails.logger.info('Starting Camtrap DP export')
+          rows_processed = 0
+
+          scientific_names = Set.new
+          audio_recordings = Set.new
+
+          deployments = DeploymentAccumulator.new(forced_timezone: @options.forced_timezone)
+
+          # Using find_each with default batch size to avoid pulling all records into memory at once.
+          included.find_each(batch_size: BATCH_SIZE) do |tagging|
+            deployment = deployments.add_or_update(tagging)
+            first_media = audio_recordings.add?(tagging.audio_event.audio_recording_id)
+
+            observation = Table::Observation.mapping(tagging, deployment)
+            scientific_names << observation.scientificName if observation.scientificName.present?
+            table_writers.observations << observation.ordered_values
+
+            if first_media
+              table_writers.media << Table::Media.mapping(tagging.audio_event.audio_recording,
+                deployment).ordered_values
+            end
+
+            rows_processed += 1
+            Rails.logger.info("Processed #{rows_processed} rows") if rows_processed % BATCH_SIZE == 0
+          end
+
+          raise ArgumentError, 'Filter returned no data, cannot export' if rows_processed.zero?
+
+          write_deployments(table_writers.deployments, deployments.values)
+          package = PackageMetadata.build(
+            deployments: deployments.values,
+            scientific_names: scientific_names,
+            options: @options
+          )
+
+          close_table_writers
+
+          validate_package(package)
+          write_datapackage_file(package)
+
+          manifest = zip_package
+
+          Rails.logger.info('Finished Camtrap DP export')
+
+          result = block.call(manifest)
+          result
+        ensure
+          cleanup
+          Bullet.enable = bullet_active if bullet_active
+        end
+
+        TableWriters = Data.define(:observations, :deployments, :media)
+
+        # @!attribute [r] package_path
+        #   @return [Pathname] path to the directory containing the unpacked Camtrap Data Package
+        # @!attribute [r] zip_path
+        #   @return [Pathname] path to the zip archive containing the Camtrap Data Package
+        # @!attribute [r] file_stats
+        #   @return [Hash<Symbol, Hash>] size and modification time for each package file
+        PackageManifest = Data.define(:package_path, :zip_path, :file_stats)
+
+        private
+
+        def validate_filter(filter)
+          message = 'Expected filter to be '
+          raise ArgumentError, "#{message}ActiveRecord::Relation" unless filter.is_a?(::ActiveRecord::Relation)
+          raise ArgumentError, "#{message}Tagging relation" unless filter.klass == Tagging
+        end
+
+        def validate_exporter_options(exporter_options)
+          return if exporter_options.is_a?(RequiredExporterOptions)
+
+          raise ArgumentError, 'exporter_options must be a RequiredExporterOptions object'
+        end
+
+        # Eager load audio_events because each audio_event usually has one tagging, so a join avoids
+        # repeated queries, and audio_event columns are mostly small scalars, so the joined rows stay cheap.
+        # Then preload recording and site to avoid duplicating their larger JSON blobs across tagging rows.
+        def included
+          @filter
+            .eager_load(:audio_event)
+            .preload(:tag, :creator, audio_event: [:provenance, { audio_recording: :site }])
+        end
+
+        # @param writer [CSV] a CSV writer for the deployments table
+        # @param deployments [Array<DeploymentAccumulator::Deployment>] the deployments to write
+        # @return [void]
+        def write_deployments(writer, deployments)
+          deployments.each do |deployment|
+            writer << Table::Deployment.mapping(deployment, user: @options.user,
+              should_obfuscate: @options.should_obfuscate).ordered_values
+          end
+        end
+
+        # @param package [Descriptor::Package] the package descriptor to write
+        # @return [void]
+        def write_datapackage_file(package)
+          package_files[:datapackage].write(JSON.pretty_generate(package.to_h))
+        end
+
+        def validate_package(package)
+          Validator.validate_package(package, package_path:)
+        end
+
+        def table_writers
+          @table_writers ||= TableWriters.new(
+            observations: open_table_csv(package_files[:observations], headers: Table::Observation.attribute_names),
+            deployments: open_table_csv(package_files[:deployments], headers: Table::Deployment.attribute_names),
+            media: open_table_csv(package_files[:media], headers: Table::Media.attribute_names)
+          )
+        end
+
+        # @return [CSV] an open CSV IO object in write-only mode
+        def open_table_csv(path, headers:)
+          CSV.open(path, 'w', write_headers: true, headers: headers, force_quotes: true)
+        end
+
+        def close_table_writers
+          return if @table_writers.nil?
+
+          @table_writers.deconstruct.each { |writer| writer.close unless writer.nil? || writer.closed? }
+        end
+
+        def exporter_temp_dir
+          @exporter_temp_dir ||= Pathname.new(Dir.mktmpdir('exporter_', BawWorkers::Config.temp_dir))
+        end
+
+        def package_path
+          @package_path ||= (exporter_temp_dir / PACKAGE_PATH).mkpath
+        end
+
+        def package_files
+          @package_files ||= PACKAGE_FILENAMES.transform_values { |filename| package_path / filename }
+        end
+
+        def zip_path
+          @zip_path ||= exporter_temp_dir / ZIP_PATH
+        end
+
+        # Create a zip archive of the package files and return a manifest of metadata about the generated export.
+        # Uses Zlib::DEFAULT_COMPRESSION which is equivalent to a 6 (valid values from 0 to 9).
+        # @return [PackageManifest]
+        def zip_package
+          Zip::File.open(zip_path, create: true) do |zip|
+            package_files.each_value { |path| zip.add(path.basename.to_s, path) }
+          end
+
+          PackageManifest.new(
+            package_path: package_path,
+            zip_path: zip_path,
+            file_stats: package_files.transform_values { |path| { size: path.size, mtime: path.mtime } }
+          )
+        end
+
+        def cleanup
+          close_table_writers
+          FileUtils.rm_rf(@exporter_temp_dir) if @exporter_temp_dir&.exist?
+        end
+      end
+    end
+  end
+end
--- a/lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/package_metadata.rb
+++ b/lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/package_metadata.rb
@@ -1,0 +1,162 @@
+# frozen_string_literal: true
+
+require_relative 'errors/camtrap_dp_error'
+
+module BawWorkers
+  module Export
+    module CamtrapDp
+      # Responsible for generating the package descriptor and associated metadata including spatial, temporal, and
+      # taxonomic coverage.
+      class PackageMetadata
+        # @param deployments [Array<DeploymentAccumulator::Deployment>] the deployments included in the package
+        # @param scientific_names [Array<String>] the scientific names included in the package
+        # @param options [ExportOptions] the export options
+        # @return [Descriptor::Package] the package descriptor
+        def self.build(deployments:, scientific_names:, options:)
+          temporal = temporal_coverage(deployments)
+          project = get_project(deployments)
+          licenses = options.emit_project_license ? project_license(project) : nil
+
+          project = Descriptor::Project.new(
+            title: project.name,
+            description: project.description,
+            path: Api::UrlHelpers.project_url(id: project.id),
+            samplingDesign: options.project_sampling_design,
+            captureMethod: options.project_capture_method,
+            individualAnimals: individual_animals,
+            observationLevel: observation_level,
+            licenses: licenses
+          )
+
+          # TODO: fill name and id fields when closed: https://github.com/QutEcoacoustics/baw-server/issues/1021
+          # TODO: add generated contributors when closed: https://github.com/QutEcoacoustics/baw-server/issues/1022
+          Descriptor::Package.new(
+            profile: Profile::PROFILE_SOURCE_URL.to_s,
+            #name:,
+            #id:,
+            created: Time.current.utc.iso8601,
+            title: options.package_title,
+            contributors: options.contributors,
+            project: project,
+            spatial: spatial_coverage(deployments, options),
+            temporal: Descriptor::Temporal.new(
+              start: temporal[:start],
+              end: temporal[:end]
+            ),
+            taxonomic: taxonomic_coverage(scientific_names),
+            sources: sources,
+            resources: [
+              BawWorkers::Export::CamtrapDp::Descriptor::Resource.new(name: 'deployments', path: DEPLOYMENTS_FILENAME.to_s,
+                schema: load_table_schema(:deployments)),
+              BawWorkers::Export::CamtrapDp::Descriptor::Resource.new(name: 'media', path: MEDIA_FILENAME.to_s,
+                schema: load_table_schema(:media)),
+              BawWorkers::Export::CamtrapDp::Descriptor::Resource.new(name: 'observations', path: OBSERVATIONS_FILENAME.to_s,
+                schema: load_table_schema(:observations))
+            ]
+          )
+        end
+
+        def self.load_table_schema(name)
+          path = Profile::DIRECTORY / Profile::ASSET_FILES[name]
+          JSON.parse(path.read, symbolize_names: true)
+        end
+
+        # Use the host name and client URL as a package source.
+        def self.sources
+          [Descriptor::Source.new(title: Settings.client.host.titlecase, path: Settings.client_routes.home_url)]
+        end
+
+        # We don't support identifying individual animals currently, so this will always be false. In future, if we
+        # add a platform feature to support this, we can make this configurable.
+        def self.individual_animals = false
+
+        # We don't distinguish currently, so default to media. In the future, we might be able to merge our 'media'
+        # observations (audio_events) into an 'event' observation level, to reduce duplication.
+        def self.observation_level = ['media']
+
+        # Return the first project in the deployments, under the assumption packages are scoped to a single project.
+        # @param deployments [Array<DeploymentAccumulator::Deployment>]
+        # @return [Project]
+        def self.get_project(deployments)
+          deployments.first&.site&.projects&.first
+        end
+
+        # Generate the licenses descriptor from the first project found
+        #
+        # In the database, existing project licenses are SPDX identifiers. But it's currently a free text field, so we
+        # will use string length as a proxy for being an identifier, and raise an error if we encounter something longer
+        # than 32 characters. We only store one license per project, so we use the same license for both required
+        # scopes: data and media.
+        def self.project_license(project)
+          project_license = project&.license
+
+          if project_license.nil?
+            nil
+          elsif project_license.length <= 32
+            [Descriptor::License.new(name: project_license, scope: 'data'),
+             Descriptor::License.new(name: project_license, scope: 'media')]
+          else
+            raise Errors::ProjectLicenseError, "Found unsupported custom license (>32 chars): #{project_license}"
+          end
+        end
+
+        # Calculate descriptor-level temporal coverage from the deployment bounds.
+        #
+        # Rails TimeWithZone comparisons are based on the UTC instant, but the offset is preserved in the output, so
+        # it's possible to have a start / end in different timezones.
+        #
+        # @param deployments [Array<DeploymentAccumulator::Deployment>]
+        # @return [Hash] with `:start` and `:end` keys containing ISO 8601 timestamps
+        def self.temporal_coverage(deployments)
+          return { start: '', end: '' } if deployments.empty?
+
+          {
+            start: deployments.map(&:start).min,
+            end: deployments.map(&:end).max
+          }
+        end
+
+        def self.spatial_coverage(deployments, options)
+          return {} if deployments.empty?
+
+          coords = deployments.map { |deployment|
+            site = deployment.site
+            lat = site.public_latitude(user: options.user, should_obfuscate: options.should_obfuscate)
+            lon = site.public_longitude(user: options.user, should_obfuscate: options.should_obfuscate)
+            [lon, lat]
+          }
+
+          return { type: 'Point', coordinates: coords.first } if coords.size == 1
+
+          lons = coords.map(&:first)
+          lats = coords.map(&:last)
+          min_lon, max_lon = lons.minmax
+          min_lat, max_lat = lats.minmax
+
+          if has_zero_area?(min_lon, max_lon, min_lat, max_lat)
+            { type: 'MultiPoint', coordinates: coords.uniq }
+          else
+            {
+              type: 'Polygon',
+              coordinates: [[
+                [min_lon, min_lat],
+                [max_lon, min_lat],
+                [max_lon, max_lat],
+                [min_lon, max_lat],
+                [min_lon, min_lat]
+              ]]
+            }
+          end
+        end
+
+        def self.has_zero_area?(min_lon, max_lon, min_lat, max_lat)
+          min_lon == max_lon || min_lat == max_lat
+        end
+
+        def self.taxonomic_coverage(scientific_names)
+          scientific_names.sort.map { |name| Descriptor::Taxonomic.new(scientificName: name) }
+        end
+      end
+    end
+  end
+end
--- a/lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/profile.rb
+++ b/lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/profile.rb
@@ -1,0 +1,142 @@
+# frozen_string_literal: true
+
+module BawWorkers
+  module Export
+    module CamtrapDp
+      module Profile
+        SOURCE_URL = 'https://raw.githubusercontent.com/camera-traps/bioacoustics/e4b4722fb453f5ca39c39ea3c4f348e9953f0084/camtrap-dp/1.0.2/'
+        ASSET_FILES = {
+          profile: 'camtrap-dp-profile-acoustic.json',
+          deployments: 'deployments-table-schema-acoustic.json',
+          media: 'media-table-schema-acoustic.json',
+          observations: 'observations-table-schema-acoustic.json'
+        }
+
+        DIRECTORY = Pathname(__dir__) / 'profile_assets'
+        README_PATH = DIRECTORY / 'README.md'
+        PROFILE_PATH = DIRECTORY / ASSET_FILES[:profile]
+        LOCAL_VALIDATION_PROFILE_PATH = PROFILE_PATH.sub_ext('.local.json')
+
+        # Used in the output package descriptor to indicate the profile used for validation.
+        PROFILE_SOURCE_URL = URI.join(SOURCE_URL, ASSET_FILES[:profile])
+
+        # References in the camtrap-dp-profile-acoustic.json to be resolved
+        EXTERNAL_SCHEMA_REFS = [
+          'http://json.schemastore.org/geojson.json',
+          'https://specs.frictionlessdata.io/schemas/data-package.json',
+          'https://geojson.org/schema/GeoJSON.json'
+        ]
+
+        # Download the Profile::ASSET_FILES into the Profile::DIRECTORY
+        # @return [Hash] manifest including the downloaded assets with their local paths.
+        def self.download
+          FileUtils.mkdir_p(DIRECTORY)
+
+          ASSET_FILES.each_with_object({}) do |(key, value), hash|
+            hash[key] = download_fixture(value)
+          end => downloaded_assets
+
+          {
+            source_url: SOURCE_URL,
+            downloaded_assets: downloaded_assets,
+            completed_at: Time.current
+          }
+        end
+
+        # Download files from Profile::SOURCE_URL into the Profile::DIRECTORY.
+        # @param name [String] the filename to download; a value in Profile::ASSET_FILES
+        # @return [String] the file path of the downloaded file relative to the application root
+        def self.download_fixture(name)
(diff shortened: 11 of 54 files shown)


app/models/permission.rb
```diff
@@ -48,6 +48,12 @@ class Permission < ApplicationRecord
     { id: l, name: Access::Core.get_level_name(l) }
   }
 
+  # Returns true when this permission grants access to anonymous or logged-in users.
+  # @return [Boolean]
+  def broad_access?
+    allow_anonymous || allow_logged_in
+  end
```
atruskie: why not name it `public_access?`

Also two notes: 

1) allow_anonymous and allow_logged_in are flags that change the permission subject between a specific user (user_id is or is not null), or logged in users, or users with id null (anonymous). 

This method as it is strictly tests the type of record, it does not test access. It should test level is above permissions level none (or level is in any of the levels that do not include none).

Granted this is a nitpick/non-issue since we don't encode none-levels in the database, but the rest of the permissions architecture does not make that assumption, and neither should this code.

2) ~~note to selves: check where this is used. The allow_anaonymous and allow_logged_in flags are set on different records, so I expect to see a `.any?(&:broad_access?)` at the call site of whereever this is used.~~ Yup found it, excellent work.

lib/gems/baw-workers/lib/baw_workers/dry/struct.rb
```diff
@@ -0,0 +1,13 @@
+module BawWorkers
+  module Dry
+    module Struct
+      # Return an array of schema values in attribute order, including nil values.
+      # Used for csv export where length and order of values must match the header row.
+      # @return [Array]
+      def full_values
```
atruskie:
```suggestion
      def ordered_values
```

?

This is a nice solution, and as long as it is covered with unit tests it's fine.

But I just wanted to point out a different solution, that is more reobust and would only be used if necessary, would be to encode the ordering somehow. E.g. we could define an `.order` extension to the type defintions, or define an `ordered_attribute` macro that stores some extra metadata when the struct attributes are defined.

lib/gems/baw-workers/lib/baw_workers/dry/types.rb
```diff
@@ -5,6 +5,20 @@ module Dry
     # custom dry-types
     # https://dry-rb.org/gems/dry-types/master/getting-started/
     module Types
+      # Convenient place to apply our dry struct extension
+      ::Dry::Struct.prepend(BawWorkers::Dry::Struct)
```
atruskie: I've done similar inadvisable things in this codebase. Perhaps this is better placed in in a patch inside of patches?

If you don't want to do patches, we could define BawWorkers::Dry::OrderedStruct instead which would be a patchless solution.

lib/gems/baw-workers/lib/baw_workers/dry/types.rb
```diff
@@ -5,6 +5,20 @@ module Dry
...
+      TimeWithPrecision = Class.new do
+        def initialize(time, precision:)
+          @time = time
+          @precision = precision
+        end
+
+        def to_s
+          @time.iso8601(@precision)
+        end
+      end
```
atruskie: I'm fairly sure this does not belong in this file. 

This is a general definition that has nothing to do with Dry.


Alternative idea: why didn't you subclass Time?

lib/gems/baw-workers/lib/baw_workers/dry/types.rb
```diff
@@ -54,6 +68,67 @@ module Types
...
+      UtcTimeSeconds = Constructor(TimeWithPrecision) { |input|
+        next nil if input.blank?
+
+        TimeWithPrecision.new(UtcTime[input], precision: 0)
+      }
+
+      # Used in the Camtrap Data Package export for timestamps serialized with microsecond precision.
+      #
+      # @return [TimeWithPrecision] wrapper that preserves the offset on Time inputs and formats via `iso8601(6)`.
+      UtcTimeMicros = Constructor(TimeWithPrecision) { |input|
```
atruskie:
```suggestion
      UtcTimeMicroseconds = Constructor(TimeWithPrecision) { |input|
```

lib/gems/baw-workers/lib/baw_workers/dry/types.rb
```diff
@@ -54,6 +68,67 @@ module Types
...
+        next nil if input.blank?
+
+        TimeWithPrecision.new(UtcTime[input], precision: 6)
+      }
+
+      # `url-or-path` is a frictionless type for a string that must either be a fully qualified URL or a relative POSIX path.
+      URLOrPath = Types::String
```
atruskie:
```suggestion
      UrlOrPath = Types::String
```

atruskie: Is it worth adding any kind of validation onto this?

lib/gems/baw-workers/lib/baw_workers/dry/types.rb
```diff
@@ -54,6 +68,67 @@ module Types
...
+      )
+
+      # package.taxonnomic[].taxonRank
+      TaxonRank = Types::String.enum(
+        'kingdom', 'phylum', 'class', 'order',
+        'family', 'genus', 'species', 'subspecies'
+      )
+
+      # Used for documentation value. The profile includes GeoJSON schema validation.
+      GeoJSON = Types::Hash
```
atruskie: Should these be isolated to a CamTrap specific Types module?

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/descriptor/contributor.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module BawWorkers
+  module Export
+    module CamtrapDp
+      # Generate automatic + merge extra
+      # TODO: all project owners should be something in the array, and all tagging creators should be as well
+      # ('contributor') - even verification creators, plus optional add more in from param
+      # Should be able to have a list of users and we can map a user to a contributor descriptor
+      # TODO: but we only have user_names, not real names, i'll need to check this.
```
atruskie: cross link to issues filed under the camtrap dp epic

atruskie: also, seems like these comments are in the wrong place. The struct is just a a simple storage mechanism. You want to leave the comment near where the struct is/would be constructed.

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/descriptor/contributor.rb
```diff
@@ -0,0 +1,21 @@
...
+      class Descriptor::Contributor < Descriptor
+        attribute :title, Types::String # name/title of the person or organisation
```
atruskie:
```suggestion
        # name/title of the person or organisation
        attribute :title, Types::String 
```

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/descriptor/contributor.rb
```diff
@@ -0,0 +1,21 @@
...
+        attribute :role, Types::Role
+
+        attribute? :email, Types::String.optional
+        attribute? :path, Types::String.optional
+        attribute? :organization, Types::String.optional
```
atruskie: I just checked the schema: these values are optional attributes, but they do not accept null. This typing is thus incorrect. Is there a reason for this?

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/descriptor/contributor.rb
```diff
@@ -0,0 +1,21 @@
...
+      # TODO: all project owners should be something in the array, and all tagging creators should be as well
+      # ('contributor') - even verification creators, plus optional add more in from param
+      # Should be able to have a list of users and we can map a user to a contributor descriptor
+      # TODO: but we only have user_names, not real names, i'll need to check this.
+      class Descriptor::Contributor < Descriptor
+        attribute :title, Types::String # name/title of the person or organisation
+        attribute :role, Types::Role
+
+        attribute? :email, Types::String.optional
+        attribute? :path, Types::String.optional
```
atruskie: Shouldn't this be your path type?

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/descriptor/license.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+module BawWorkers
+  module Export
+    module CamtrapDp
+      class Descriptor
+        class License < Descriptor
+          attribute :name, Types::String
+          attribute? :path, Types::String.optional
```
atruskie: shouldn't this be your path type?

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/descriptor/package.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+# TODO: sources field : open ecoacoustics
+#   we have the name and title in settings file
+#   ecosounds.org
```
atruskie: wdym?

Bur more to the point, tracking issue and cross link

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/profile.rb
```diff
@@ -0,0 +1,112 @@
...
+          profile: 'camtrap-dp-profile-acoustic.json',
+          deployments: 'deployments-table-schema-acoustic.json',
+          media: 'media-table-schema-acoustic.json',
+          observations: 'observations-table-schema-acoustic.json'
+        }
+
+        DIRECTORY = File.join(__dir__, 'profile_assets')
+        README_PATH = File.join(DIRECTORY, 'README.md')
+        PROFILE_PATH = File.join(DIRECTORY, ASSET_FILES[:profile])
+        LOCAL_VALIDATION_PROFILE_PATH = PROFILE_PATH.gsub('.json', '.local.json')
```
atruskie:
```suggestion
        DIRECTORY = Pathname(__dir__) / 'profile_assets'
        README_PATH = DIRECTORY / 'README.md'
        PROFILE_PATH = DIRECTORY / ASSET_FILES[:profile]
        LOCAL_VALIDATION_PROFILE_PATH = PROFILE_PATH.sub_ext('.local.json')
```

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/descriptor/resource.rb
```diff
@@ -0,0 +1,31 @@
...
+          end
+
+          def self.load_schema(path) = JSON.parse(File.read(path), symbolize_names: true)
+
+          DEFAULT_RESOURCES = [
+            new(name: 'deployments', path: DEPLOYMENTS_FILENAME, schema: load_table_schema(:deployments)),
+            new(name: 'media', path: MEDIA_FILENAME, schema: load_table_schema(:media)),
+            new(name: 'observations', path: OBSERVATIONS_FILENAME, schema: load_table_schema(:observations))
+          ].freeze
+        end
```
atruskie: OK, there are a couple of issues here:

1. Is `load_schema` used anywhere else? Why not inline it?
2. Use Pathname. e.g. `Profile::Directory / Profile::ASSET_FILES[:media]` and `path.read`
3. Where are those `_FILENAME` constants coming from?
4. If you just make these load schema methods work on a class instance, you can simply do `new(name: 'media')` and you can set path (i think) and schema in the constructor
5. But most importantly: what is this struct for? These strcuts are best used to represent a type you have to seraialize too. They're the destination for your data after marshalling. Adding in schema here makes this structu look operational; in which case it shouldn't be in this folder with all the other structs.

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/descriptor/source.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+module BawWorkers
+  module Export
+    module CamtrapDp
+      class Descriptor
+        class Source < Descriptor
```
atruskie: A short title would be useful here, source is super ambiguous to the casual reader

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/descriptor/temporal.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module BawWorkers
+  module Export
+    module CamtrapDp
+      class Descriptor
+        class Temporal < Descriptor
+          attribute :start, Types::String
+          attribute :end, Types::String
```
atruskie: Why aren't these one of our date types?

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/profile_assets/README.md
```diff
@@ -0,0 +1,25 @@
+# Downloaded profile assets:
```
atruskie:
```suggestion
# CamTrap DP Bioacoustics Downloaded profile assets

The files in this directory (including this readme) are generated with the `baw:camtrap_dp:update` rake task. Do not edit these files manually,

Generation date:
Source: <repo>#<hash>
```

Obviously update your rake task

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/profile_assets/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+# Created local validation profile:
+
+{local_validation_profile:
+  "lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/profile_assets/camtrap-dp-profile-acoustic.local.json",
+ external_references_inlined:
+  {"https://specs.frictionlessdata.io/schemas/data-package.json" => true,
+   "http://json.schemastore.org/geojson.json" => true,
+   "https://geojson.org/schema/GeoJSON.json" => true},
+ completed_at: 2026-06-26 03:30:47.548083417 +0000}
```
atruskie: Image

We could neaten this up a bit yeah?

use a bulelt list, use relative paths for the profile assets, don't jsut dump a hash

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/table/deployment.rb
```diff
@@ -0,0 +1,115 @@
...
+          attribute? :locationID, Types::Coercible::String.optional
+
+          # Name given to the deployment location.
+          attribute? :locationName, Types::String.optional
+
+          # Latitude of the deployment location in decimal degrees, using the WGS84 datum.
+          attribute :latitude, Types::Nominal::Decimal
+
+          # Longitude of the deployment location in decimal degrees, using the WGS84 datum.
+          attribute :longitude, Types::Nominal::Decimal
```
atruskie, line 26: constrain values?

Also these is a useful enough definitions to be stored elsewhere.

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/table/deployment.rb
```diff
@@ -0,0 +1,115 @@
+module BawWorkers
+  module Export
+    module CamtrapDp
+      module Table
+        # Represents a row in the `deployment` table of the Camtrap Data Package.
+        #
+        # Attributes are defined in schema order (required for CSV validation).
+        class Deployment < ::Dry::Struct
```
atruskie: As I understand it there should be no `attribute?` definitions in this class - all fields are required for tables, no?

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/table/deployment.rb
```diff
@@ -0,0 +1,115 @@
...
+              'culvert', 'burrow', 'nestSite', 'carcass', 'waterSource', 'fruitingTree').optional
+
+          # Short characterization of the habitat at the deployment location.
+          attribute? :habitat, Types::String.optional
+
+          # Deployment group(s) associated with the deployment. Deployment groups can have a spatial (arrays, grids, clusters), temporal (sessions, seasons, months, years) or other context. Formatted as a pipe (`|`) separated list for multiple values, with values preferably formatted as `key:value` pairs.
+          attribute? :deploymentGroups, Types::String.optional
+
+          # Tag(s) associated with the deployment. Formatted as a pipe (`|`) separated list for multiple values, with values optionally formatted as `key:value` pairs.
+          attribute? :deploymentTags, Types::String.optional
```
atruskie: if we ere to actually use any of these we would probably make a `PipeDelimittedList` type or something like that.

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/table/deployment.rb
```diff
@@ -0,0 +1,115 @@
...
+            site = deployment.site
+
+            # Because we are treating site as a single deployment, site id is used for both the deploymentID and locationID.
+            Deployment.new(
+              deploymentID: site.id,
+              locationID: site.id,
+              locationName: site.name,
+              latitude: (should_obfuscate ? site.obfuscated_latitude : site.latitude),
+              longitude: (should_obfuscate ? site.obfuscated_longitude : site.longitude),
+              coordinateUncertainty: nil,
```
atruskie: this defintiely needs to be filled in

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/table/deployment.rb
```diff
@@ -0,0 +1,115 @@
...
+          #   whose `export_time` method applies the forced UTC offset, site timezone, or UTC fallback.
+          # @param should_obfuscate [Boolean] whether to obfuscate the latitude and longitude values for the deployment
+          # @return [Deployment] the deployment struct with the mapped values
+          def self.mapping(deployment, should_obfuscate:)
+            site = deployment.site
+
+            # Because we are treating site as a single deployment, site id is used for both the deploymentID and locationID.
+            Deployment.new(
+              deploymentID: site.id,
+              locationID: site.id,
```
atruskie: These identifiers should be globally unique IMO. An integer is just too vague.

It should be `ecosounds.org/sites/123`. open question as to whether it should use `api` or `www` as a prefix

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/table/media.rb
```diff
@@ -0,0 +1,67 @@
...
+          attribute? :favorite, Types::Bool.optional
+          attribute? :mediaComments, Types::String.optional
+
+          # @param audio_recording [AudioRecording] the audio recording to map
+          # @param deployment [DeploymentAccumulator::Deployment] the deployment metadata for the audio recording;
+          #   its `export_time` method applies the forced UTC offset, site timezone, or UTC fallback.
+          # @return [Media] the media struct with the mapped values
+          def self.mapping(audio_recording, deployment)
+            # Manually set the site association cache on the audio recording to avoid a database query when calling `friendly_name`.
+            audio_recording = audio_recording.tap { |ar| ar.association(:site).target = deployment.site }
```
atruskie: seems like this belongs outside of this class - it has nothing to do with the emdia entry.

Also, why not .includes(:site)?

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/table/media.rb
```diff
@@ -0,0 +1,67 @@
...
+
+            file_path = Api::UrlHelpers::Base.new.audio_recording_media_original_url(audio_recording_id: audio_recording.id)
```
atruskie:
```suggestion
            file_path = Api::UrlHelpers.audio_recording_media_original_url(audio_recording_id: audio_recording.id)
```

The method missing stuff on the class instantiates the url helpers once and proxies method calls to the instance

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/table/media.rb
```diff
@@ -0,0 +1,67 @@
...
+
+            # ? Is this calculation correct? There are many cases in the database where this calculation doesn't result
+            # ? in their valid enum values. And in that case, we just have to return nil?
+            # ? Alternatively, we also emit our own bit_rate_bps field?
+            bit_depth = (audio_recording.bit_rate_bps.to_f / (audio_recording.channels * audio_recording.sample_rate_hertz))
+            bit_depth = nil unless bit_depth.in?([8, 16, 24, 32])
```
atruskie: I'm honestly shocked we don't have bit depth in our database. Your calculation is correct for (most) WAVE files, but will not be correct for MP3 or FLAC.

I don't think you should estimate it here. Rather let's add an issue to extract it when harvesting and then fix up this entry later.

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/table/media.rb
```diff
@@ -0,0 +1,67 @@
...
+              filePublic: deployment.file_public,
+              fileName: audio_recording.friendly_name,
+              fileMediatype: audio_recording.media_type,
+              exifData: nil,
+              bitDepth: bit_depth,
+              samplingFrequency: audio_recording.sample_rate_hertz,
+              gain: nil,
+              channels: audio_recording.channels,
+              favorite: nil,
+              mediaComments: audio_recording.notes.to_s
```
atruskie: I don't think we want to export notes here.  The notes we have are mostly operational junk.

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/table/observation.rb
```diff
@@ -0,0 +1,223 @@
...
+  module Export
+    module CamtrapDp
+      module Table
+        # Represents a row in the `observations` table of the Camtrap Data Package.
+        #
+        # Attributes are defined in schema order (required for CSV validation).
+        class Observation < ::Dry::Struct
+          Types = BawWorkers::Dry::Types
+
+          # Fixed: we only emit interval level observations
```
atruskie: Why? I remember you had to explain it to me and I am wondering again.

atruskie: Ok I see an explanation only below. Maybe thus just needs to be something like: `Our concept of an event maps to the interval type of observation. See explanation on #observationLevel`.

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/table/observation.rb
```diff
@@ -0,0 +1,223 @@
...
+          attribute :eventStart, Types::UtcTimeMicros
+
+          # Date and time at which the event ended. Formatted as an ISO 8601 string with timezone designator (`YYYY-MM-DDThh:mm:ssZ` or `YYYY-MM-DDThh:mm:ss¬±hh:mm`).
+          attribute :eventEnd, Types::UtcTimeMicros
+
+          # Level at which the observation was classified. `media` for media-based observations that are directly associated with a media file (`mediaID`). These are especially useful for machine learning and don't need to be mutually exclusive (e.g. multiple classifications are allowed). `event` for event-based observations that consider an event (comprising a collection of media files). These are especially useful for ecological research and should be mutually exclusive, so that their `count` can be summed. Acoustic extension adds 'interval' to accepted enum values but seemingly only to the field on observations table, not the package level observationLevel field
+          attribute :observationLevel, Types::String.enum('media', 'event', 'interval')
```
atruskie, line 35: I'm need an ELI5 here.

- media: a classification of a whole file. Makes sense for short media or images
- event: a collection of media observations that can be grouped together. Makes sense for a series of images that contain the same classified individual. (I think I got this wrong again)
- interval: a bioacoustics adaptation that allows an event to occur within media by having start and end times

?

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/table/observation.rb
```diff
@@ -0,0 +1,223 @@
...
+
+          # Fields not covered by the standard that we would like to include in the export:
+          #
+          # audio_event.score - The score is unbounded and not a probability, so it cannot be used for classificationProbability.
+          # verifications - e.g. verification_correct count, ratio of correct to total verifications etc.#
+          #
+          # See Camtrap DP FAQ for ways to include additional fields https://camtrap-dp.tdwg.org/faq/#measurements;
+          # either as key:value pairs in the observationTags field, or as a custom table with a schema.  Also see the
+          # community feedback issue from the bioacoustic extension report https://github.com/tdwg/camtrap-dp/issues/457
+          #
```
atruskie, line 166: Good notes. We have a strong preference for just adding an extra column (piped kv pairs are annoying, as is stitching together an a 1:1 table from another file.

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/table/observation.rb
```diff
@@ -0,0 +1,223 @@
...
+          # @param tagging [Tagging] the tagging to map
+          # @param deployment [DeploymentAccumulator::Deployment] the deployment metadata for the tagging; its
+          #   `export_time` method applies the forced UTC offset, site timezone, or UTC fallback.
+          # @return [Observation] the observation struct with the mapped values
+          def self.mapping(tagging, deployment)
+            ae = tagging.audio_event
+            ar = ae.audio_recording
+
+            # ? Is this future proof?
+            classification_method = ae.provenance_id ? 'machine' : nil
```
atruskie: Great question. Currently we make an assumption that provenances are all algorithms. But we've had feedback that we want to track human effort to for manual imports, so provenance might be widened.  

Here's an idea: define `machine_generated?` (or similar) on `Provenance` and have it return `true` always. If we ever widen the provenance concept, we can insert more completed logic, and it will be forwards compatible with any existing uses.

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/table/observation.rb
```diff
@@ -0,0 +1,223 @@
...
+
+            # ? When audio_event.provenance is null, `tagging.creator` queries the user table:
+            # - Should we eager load the user association to avoid this?
+            # - The user_name is not a real name, is that an issue for the package or for ALA?
```
atruskie: In a similar vein: define `User::full_name` which produces a full name if the information is available, otherwise falls back to `user_name`.

Collate questions we have for ALA - they might not use this field at all,
it seems if anything, this string should match an entry from the collaborators entry if anything.

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/table/observation.rb
```diff
@@ -0,0 +1,223 @@
...
+
+            observation_type = observation_type(tagging.tag)
+
+            Observation.new(
+              observationID: tagging.id,
+              deploymentID: ar.site_id,
+              mediaID: ar.id,
+              eventID: nil,
+              eventStart: deployment.export_time(ar.recorded_date + ae.start_time_seconds.seconds),
+              eventEnd: deployment.export_time(ar.recorded_date + ae.end_time_seconds.seconds),
```
atruskie: I don't know what export time does, but that seems fishy. You made these fancy types above to control precision... what else do you need to do? We shall find out 🕵️‍♂️

Also use/define `start_date` and `end_date` on `AudioRecording`

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/deployment_accumulator.rb
```diff
@@ -0,0 +1,92 @@
...
+        # @!attribute [r] offset_or_timezone
+        #   @return [UTCOffsetString, TZInfo::Timezone, ActiveSupport::TimeZone] to convert stored UTC timestamps into
+        #   the export representation for the deployment.
+        Deployment = Data.define(:site, :start, :end, :file_public, :offset_or_timezone) {
+          # Convert a stored UTC instant into the timestamp representation used by the export.
+          #
+          # @param time [Time, ActiveSupport::TimeWithZone, nil] timestamp to export.
+          # @return [Time, ActiveSupport::TimeWithZone, nil] timestamp converted to UTC, a fixed offset, or a site
+          #   timezone, ready for Camtrap-DP ISO 8601 serialization.
+          def export_time(time)
```
atruskie: I have mixed feeling about this. I think the name is the biggest issue. Suggest `ensure_timezone` or `time_in_zone` or something like that. 

>  \# Convert a stored UTC instant into the timestamp representation used by the export.

Seems misleading. More like "Ensures all times in a deployment are in a uniform Timezone for exporting".

I understand now why you have this here. I don't think I would have hung this off of deployment (it is a logical place to collect the information, a little weirder to read about it when seeing it used though). 

I think I would have made models return things like `startDate` in timezones automatically, and only added the override on the serialization step at the end.

I think with the name change suggested above it's more self-describing and fine as is.

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/deployment_accumulator.rb
```diff
@@ -0,0 +1,92 @@
...
+        def initialize(force_utc_offset: nil)
+          @deployments = {}
+
+          # Store forced offsets as formatted strings so conversion can use the same stable offset for every timestamp,
+          # independent of date-specific daylight saving rules.
+          @force_utc_offset = force_utc_offset ? UTCOffsetString.new(ActiveSupport::TimeZone.seconds_to_utc_offset(force_utc_offset)) : nil
+        end
+
+        # +HH:MM formatted UTC offset string.
+        UTCOffsetString = Data.define(:utc_offset)
```
atruskie: why do we need a data for one field?

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/deployment_accumulator.rb
```diff
@@ -0,0 +1,92 @@
...
+      # recordings at the site in the result set.
+      class DeploymentAccumulator
+        # @param force_utc_offset [Integer, nil] optional UTC offset in seconds to use instead of site timezones for
+        #   all deployments accumulated by this instance.
+        def initialize(force_utc_offset: nil)
+          @deployments = {}
+
+          # Store forced offsets as formatted strings so conversion can use the same stable offset for every timestamp,
+          # independent of date-specific daylight saving rules.
+          @force_utc_offset = force_utc_offset ? UTCOffsetString.new(ActiveSupport::TimeZone.seconds_to_utc_offset(force_utc_offset)) : nil
```
atruskie: So, just type check this, ensure it is a TimeZone type (active support or tzinfo). Then you don't need to accept seconds and parse to a string.

You can make special custom timezones from offsets like this: `ActiveSupport::TimeZone.create('Fixed', 600, TZInfo::Timezone.get('GMT'))`. Now the caller has to provide a timezone object and this code gets much simpler.

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/deployment_accumulator.rb
```diff
@@ -0,0 +1,92 @@
...
+          # @param time [Time, ActiveSupport::TimeWithZone, nil] timestamp to export.
+          # @return [Time, ActiveSupport::TimeWithZone, nil] timestamp converted to UTC, a fixed offset, or a site
+          #   timezone, ready for Camtrap-DP ISO 8601 serialization.
+          def export_time(time)
+            return nil if time.blank?
+            return time.utc if offset_or_timezone.nil?
+
+            # `getlocal` preserves the instant and renders it with the requested fixed offset.
+            return time.getlocal(offset_or_timezone.utc_offset) if offset_or_timezone.is_a?(UTCOffsetString)
+
```
atruskie: apply previous suggestion and you don't need this

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/deployment_accumulator.rb
```diff
@@ -0,0 +1,92 @@
...
+            end: nil,
+            file_public: audio_recording.site.public_site?,
+            offset_or_timezone: @force_utc_offset || site_timezone(audio_recording.site)
+          )
+
+          @deployments[audio_recording.site_id] = deployment.with(
+            start: deployment.export_time(earliest(deployment.start, audio_recording.recorded_date)),
+            end: deployment.export_time(latest(deployment.end, audio_recording.recorded_end_date))
+          )
+        end
```
atruskie, line 62: Why not just use the first and last recording from each site?

I guess this is more stable if recordings frow? Makes more sense for filtered results?
But it's a pretty lackluster representation of a site or a deployment?

Does the spec offer guidance?

Why are we running export_time on these dates? In other places you've used export time when you create the struct.... seems inconsistent.

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/deployment_accumulator.rb
```diff
@@ -0,0 +1,92 @@
...
+        def latest(*times) = times.compact.max
+
+        def values = @deployments.values
+
+        private
+
+        # @param site [Site] site whose timezone should be used for export timestamps.
+        # @return [TZInfo::Timezone, ActiveSupport::TimeZone] timezone of site or UTC fallback for nil or zero-offset timezones.
+        def site_timezone(site)
+          return ActiveSupport::TimeZone['UTC'] if site.timezone.blank? || site.timezone[:utc_total_offset].zero?
```
atruskie: Why handle 0 offset timezones specially at all?

Note I think this is actually a mistake. GMT has an offset of zero and distinct from UTC.

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/descriptor.rb
```diff
@@ -0,0 +1,24 @@
...
+
+        # Override the default to_h method to remove nil values with compact.
+        #
+        # We use explicit nil values for optional fields for clarity and maintainability, but null values in the package
+        # output raise type errors during validation unless the type allows it.
+        def to_h
+          self.class.schema.each_with_object({}) do |key, result|
+            result[key.name] = ::Dry::Struct::Hashify[self[key.name]] if attributes.key?(key.name)
+            result.compact!
+          end
```
atruskie: This customization may no longer be needed if the suggested `attribute[?]` changes are applied

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/exporter.rb
```diff
@@ -0,0 +1,182 @@
...
+        #   @return [Integer, nil] optional UTC offset in seconds to apply to every exported recording, event,
+        #     deployment, and temporal coverage timestamp. When omitted, each deployment uses its site's timezone;
+        #     sites with no timezone or a zero UTC offset export in UTC.
+        RequiredExporterOptions = Data.define(
+          :should_obfuscate,
+          :contributors,
+          :project_capture_method,
+          :project_sampling_design,
+          :emit_project_license,
+          :force_utc_offset
```
atruskie: suggestion: `force_tiemzone`

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/exporter.rb
```diff
@@ -0,0 +1,182 @@
...
+        )
+
+        def initialize(filter, **exporter_options)
+          @filter = validate_filter(filter)
+          @options = required_exporter_options(exporter_options)
```
atruskie: The point of using a Data like `RequiredExporterOptions` is to force the consumer to provide valid options you don't have to check. If you just allow passing a hash/keywords then you're just doing double duty.

Either just have the keywords as normal (and the documentation for kewords args),  or required a single second argument of `RequiredExporterOptions`

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/exporter.rb
```diff
@@ -0,0 +1,182 @@
...
+        end
+
+        # Generates the export and yields a manifest of metadata about the generated export to the provided block.
+        #
+        # @yield [PackageManifest] export metadata, including the path to the package zip file
+        # @return the result of the block
+        def call(&block)
```
atruskie, line 79: This is a potentially long running operation. We need some logging in here

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/exporter.rb
```diff
@@ -0,0 +1,182 @@
...
+        TableWriters = Data.define(:observations, :deployments, :media)
+        PackageManifest = Data.define(:package_path, :zip_path, :file_stats)
+
+        private
+
+        def validate_filter(filter)
+          message = 'Expected filter to be '
+          raise ArgumentError, "#{message}ActiveRecord::Relation" unless filter.is_a?(::ActiveRecord::Relation)
+          raise ArgumentError, "#{message}Tagging relation" unless filter.klass == Tagging
+          raise ArgumentError, 'Filter returned no data, cannot export' if filter.empty?
```
atruskie: This is either doing one of two thins and thy're both bad:

If `empty?` is evaluating the query, then we'll run the filter query twice: once to a count(*) = = (empty check) and twice in find_in_batches. It's not going trash the ruby process but the database will still execute the core query twice.

If it's just checking id it is not null, then, it is a redundant condition with a bad message.

Either way, remove.

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/exporter.rb
```diff
@@ -0,0 +1,182 @@
...
+
+          filter
```
atruskie: nit: no point in returning a value unless you modify it

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/exporter.rb
```diff
@@ -0,0 +1,182 @@
...
+        end
+
+        def required_exporter_options(exporter_options)
+          missing = RequiredExporterOptions.members - exporter_options.keys
+          raise ArgumentError, "Missing required exporter option: #{missing.first}" if missing.any?
+
+          RequiredExporterOptions.new(**exporter_options)
+        end
```
atruskie: I mean Data does this anyway (admittedly with a less explicit error message). Why repeat it?

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/exporter.rb
```diff
@@ -0,0 +1,182 @@
...
+
+        def included
+          @included ||= @filter.includes(:tag, audio_event: [:provenance, :audio_recording])
```
atruskie: there is little reason to memoize this. It's modifying the query only, not executing anything. Performance gains are minimal and caching bugs are always hard to track down

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/exporter.rb
```diff
@@ -0,0 +1,182 @@
...
+
+        def write_deployments(writer, deployments)
+          deployments.each do |deployment|
+            writer << Table::Deployment.mapping(deployment,
+              should_obfuscate: @options.should_obfuscate).full_values
+          end
+        end
+
+        def write_datapackage_file(package)
+          File.write(package_files[:datapackage], JSON.pretty_generate(package.to_h))
```
atruskie: https://docs.ruby-lang.org/en/3.0/Pathname.html#method-i-write

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/exporter.rb
```diff
@@ -0,0 +1,182 @@
...
+        end
+
+        def close_table_writers
+          return if table_writers.nil?
+
+          table_writers.deconstruct.each { |writer| writer.close unless writer.nil? || writer.closed? }
+        end
+
+        def exporter_temp_dir
+          @exporter_temp_dir ||= Pathname.new(Dir.mktmpdir('exporter_', Pathname.new(BawWorkers::Config.temp_dir)))
```
atruskie:
```suggestion
          @exporter_temp_dir ||= Pathname(Dir.mktmpdir('exporter_', BawWorkers::Config.temp_dir))
```

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/exporter.rb
```diff
@@ -0,0 +1,182 @@
...
+        def package_path
+          @package_path ||= exporter_temp_dir.join(PACKAGE_PATH).tap { |path| FileUtils.mkdir_p(path) }
+        end
+
+        def package_files
+          @package_files ||= PACKAGE_FILENAMES.transform_values { |filename| package_path.join(filename) }
+        end
+
+        def zip_path
+          @zip_path ||= exporter_temp_dir.join(ZIP_PATH)
```
atruskie: Use pathname's and `/` to join. You can also then use `.mkpath` from Pathname to ensure the directory exists (avoids to .tap)

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/exporter.rb
```diff
@@ -0,0 +1,182 @@
...
+        end
+
+        # @return [PackageManifest]
+        def zip_package
```
atruskie, line 229: out of interest: what is the deault compression level on this zip? We would want to make sure at least some comression is enabled.

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/package_metadata.rb
```diff
@@ -0,0 +1,140 @@
...
+      # taxonomic coverage.
+      class PackageMetadata
+        def self.build(deployments:, scientific_names:, options:)
+          temporal = temporal_coverage(deployments)
+          project = get_project(deployments)
+          licenses = options.emit_project_license ? project_license(project) : nil
+
+          project = Descriptor::Project.new(
+            title: project.name,
+            description: project.description,
```
atruskie, line 22: is it valid to emit markdown?

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/package_metadata.rb
```diff
@@ -0,0 +1,140 @@
...
+            path: nil,
```
atruskie: Not the project url?

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/package_metadata.rb
```diff
@@ -0,0 +1,140 @@
...
+            captureMethod: options.project_capture_method,
+            individualAnimals: individual_animals,
+            observationLevel: observation_level,
+            licenses: licenses
+          )
+
+          Descriptor::Package.new(
+            profile: Profile::PROFILE_PATH,
+            name: nil,
+            id: nil,
```
atruskie: Seems like bothof these should be non nil ?https://specs.frictionlessdata.io/data-package/#metadata

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/package_metadata.rb
```diff
@@ -0,0 +1,140 @@
...
+            temporal: Descriptor::Temporal.new(
+              start: temporal[:start],
+              end: temporal[:end]
+            ),
+            taxonomic: taxonomic_coverage(scientific_names)
+          )
+        end
+
+        # Optional title for this specific package, different from the project title
+        def self.package_title = nil
```
atruskie: This should be an option the module

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/package_metadata.rb
```diff
@@ -0,0 +1,140 @@
...
+          }
+        end
+
+        def self.spatial_coverage(deployments, options)
+          return {} if deployments.empty?
+
+          coords = deployments.map { |deployment|
+            site = deployment.site
+            lat = (options.should_obfuscate ? site.obfuscated_latitude : site.latitude).to_f
+            lon = (options.should_obfuscate ? site.obfuscated_longitude : site.longitude).to_f
```
atruskie: use `Site#public_[latitude|longitude]`.

I missed an example of this elsewhere too. Probably in the deployments struct?

atruskie: Does the to_f do anything?

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/profile.rb
atruskie, line 1: generally use pathname everywhere you can

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/validator.rb
```diff
@@ -0,0 +1,80 @@
...
+          true
+        end
+
+        # Create a DataPackage::Package with the `profile` path set to our local validation profile path.
+        # See {CamtrapDp::Profile} for more information on the local validation profile.
+        #
+        # @raise [Errors::PackageLoadError] if the package cannot be loaded.
+        # @return [DataPackage::Package]
+        def self.load_package(package, package_path:)
+          package_descriptor = package.to_h.deep_stringify_keys
```
atruskie: deep_stringify should be an option of last resort. It will be relatively slow to do. I assume it is needed here?

lib/gems/baw-workers/lib/patches/data_package.rb
```diff
@@ -0,0 +1,107 @@
+module Baw
+  module DataPackage
+    module Profile
+      # Our output packages use a GitHub URI for the camtrap-dp-acoustic profile. We want to
+      # resolve this locally rather than fetch it over HTTP, but the gem only supports loading
+      # profiles from HTTP URIs or its internal registry and local file paths aren't handled.
+      # This patch adds a local file check to Profile#get_profile_from_registry before
+      # falling back to the registry lookup.
+      def get_profile_from_registry(descriptor)
```
atruskie, line 10: cross link to pull request so we know when to remove this patch

lib/gems/baw-workers/lib/patches/data_package.rb
```diff
@@ -0,0 +1,107 @@
...
+        end
+      end
+    end
+
+    # `check_pattern` calls `@value.to_json` before passing it to the regex matcher. For string values, `to_json` wraps
+    # the output in double quotes, so `audio/mpeg` becomes `"audio/mpeg"`. A pattern anchored with `^` then fails
+    # because the leading `"` sits before the expected match position.
+    # Instead, if the value is a string, pass it directly to the matcher, otherwise call `to_json` as before to preserve
+    # the original behaviour for non-string values.
+    module Constraints
```
atruskie, line 79: cross  link with issue

atruskie (changes requested):
Wow a massive amount of work. Well done!

Ok the `attribute? xxx Types::Xxx.optional` pattern occurs several times. I gather you've done it on purpose, but it does not match the schema you're writing against (Data package schema - which tends to not allow null, but instead allows attributes to be missing if they're optional).  So why have you chosen this pattern?
For the tables: they should all be `attribute` and we should leave a comment explaining why they're required even though the schema says some are not requried.

It would be great if these descriptor classes could point to the bits of the schema they implement. It would make it easy for AI tools to validate their correctness.

As a general style, we do not do end of line comments.

I think all of our ID columns should be globally unique. A common way to do this is scope them to our domain name. e.g.

- reverse domain name: org.ecosounds.sites.123
- some other format: ecosounds.org(site:123)
- or just a url that maps to a resource: ecosounds.org/sites/123 (this is my preferred choice.

Globally unique identifiers make our identifiers actually unique when that are incorporated into mix use datasets or repositories.

lib/gems/baw-workers/lib/baw_workers/dry/struct.rb
andrew-1234: Added test

lib/gems/baw-workers/lib/baw_workers/dry/types.rb
andrew-1234: Nice, changed from a mixin to a BawWorkers::Dry::OrderedStruct class that the tables now subclass

lib/gems/baw-workers/lib/baw_workers/dry/types.rb
andrew-1234: ✅ Moved to camtrap dp namespace 
That's something I didn't think of.  to be able to serialise with different precisions, just having the simple wrapper feels pretty neat, i think. and then the class doesn't need to fully behave like a Time or be a drop in replacement for a Time, which is nice?

atruskie: I'd find a wrapper annoying because you have to unwrap the value to use it. e,.g. ` date =TimeWithPrecision.new(date.value +1.days, precision: date.precision)` instead of `date = date + 1.days`.

But if you don't have to modify it then it's a non issue and a wrapper is simpler as you said.

I think the best solution is a serializer that customises output for the use case. But even that solution needs the values to that are formatted differently to be tagged. An extension to the dry typing system could store the metadata for serialization too. I think that's really the galaxy brain move.

e.g. `attribute :createdTime, Dry::Types::Time.with_precision(9)`

Then we'd have a Dry CSV serializer that checks the metadata and Serializes values, much how lke we do it for the API CSV serializer: https://github.com/QutEcoacoustics/baw-server/blob/a31be95393f8cb21cac2804c51fb95443be69be5/app/modules/api/csv.rb#L19-L30

Not necessary right now though.

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/validator.rb
andrew-1234: yea, for the datapackage gem, it uses string keys

atruskie: yeah i guess, if we were reading this back from the json file we'd just do stringify_keys: false.

But that's not the case

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/descriptor/contributor.rb
andrew-1234: it's different, this one is just a url not a url-or-path

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/descriptor/license.rb
andrew-1234: yes, fixed

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/descriptor/contributor.rb
atruskie: Could add a URL type to dry types for clarity?

```
Url= Types::String.constructor { |input| URI.parse(input) }
```

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/table/deployment.rb
andrew-1234: the uncertainty? i didn't see a field in the site table where that's stored

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/package_metadata.rb
andrew-1234: so like this right? still able to force obfuscate via the options

 (options.should_obfuscate ? site.obfuscated_latitude : site.public_latitude)

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/table/media.rb
andrew-1234: > Also, why not .includes(:site)?

do you mean why not for the whole query? because i think originally i had includes :site but we talked and you said it might be better without it, because of the query size / number of columns returned, since number of sites will be far fewer than the number of rows. something like that. 

but if you mean for audio_recording.includes(:site) it doesn't  work on an instance . maybe theres a better way though

and yea i can move it to more appropriate place

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/table/deployment.rb
andrew-1234: do you mean stored elsewhere as a type?

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/profile_assets/README.md
andrew-1234: ok sounds good

 relative to what? i made them relative to the workspace root

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/descriptor/temporal.rb
andrew-1234: fixed

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/deployment_accumulator.rb
andrew-1234: like use the first and last recording from a site, even if it's outside the filter? Yeah I think when we talked about this, when we said to get the deployment start end from audio recordings i just assumed it would be from the audio recordings in the filter only.  So I didn't consider that way. 

But I don't know if it matters much, either way it isn't a deployment. Maybe the overall site start is more interesting

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/table/deployment.rb
atruskie: if it is obfuscated, it has uncertainty. See https://github.com/QutEcoacoustics/baw-server/blob/a31be95393f8cb21cac2804c51fb95443be69be5/app/models/site.rb#L76-L78

Trickier is if `custom_obfuscated_location` then we can't back fill. We have to calculate it?

I don't actually care about the exact uncertainty being represented. I do care whether or not we flag coordinates as accurate or not. 

Custom fields don't seem to be possible, so I guess we have to fill uncertainty

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/package_metadata.rb
atruskie: OK fair point, the `public_` methods check for permission themselves, rather than allowing an override.

Maybe the public methods should accept an override? (`def public_longitude(should_obfuscate:`).

My main criticism here was the duplication of logic. The model should be in source of truth for whether coordinates were obfuscated or not. The logic should not be spread out in other modules.

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/table/media.rb
atruskie: I think I am the source of some misinformation. 

https://www.keypup.io/blog/complex-preloading-strategies-in-rails-using-custom-active-record-scopes/
https://www.ironin.it/blog/how-activerecord-decides-when-preload-and-when-eager-load-records.html

We want the preload strategy. Includes can choose  preload or eager_load, we want to force preload.

I saw this gem which could be ideal: https://github.com/DmitryTsepelev/ar_lazy_preload

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/table/deployment.rb
atruskie: no i mean valid values for the type are e.g. >-180 and <180

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/profile_assets/README.md
atruskie: to the aseets directory

e.g. `./camtrap-dp-profile-acoustic.local.json` instead of `lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/profile_assets/camtrap-dp-profile-acoustic.local.json,`

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/deployment_accumulator.rb
atruskie: talked out of band: I think our fake deployments having stable start and end times that don't change with filter. It'll just make data look consistent

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/exporter.rb
andrew-1234: for the better error message

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/exporter.rb
andrew-1234: i'm assigning the result so have to return filter here or else the return value is the message

atruskie: So, yeah still a nit, but think how a method signals intent when you use it.

e.g.

`a = my_func(a)` implies some modification to `a` may happen

`my_funct(a)` implies a side-effect has happened 

These are not guarantees. There are plenty of exceptions. But my first assumption  when reading `a = my_func(a)` is that it may have changed something.

And you'r eonly returning it so you can assign it. That's not an argument. Just have two statements.

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/exporter.rb
andrew-1234: Zlib::DEFAULT_COMPRESSION which is equivalent to a 6, and valid values are 0 to 9.

Seems to be working :
 
52K     /tmp/exporter_20260720-530453-knccn1/dp
8.0K   /tmp/exporter_20260720-530453-knccn1/dp.zip

https://github.com/rubyzip/rubyzip#compression-level

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/package_metadata.rb
andrew-1234: they aren't required which is why i left them for now

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/package_metadata.rb
andrew-1234: yea

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/package_metadata.rb
atruskie: ID will make more sense when we start generating these. It will be an ID of an export row.

Name is something we should tackle - maybe as a todo though.
We'll also have to generate a name when downloading the packages, so we'd use it in both spots.

Make issues and cross link.

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/exporter.rb
andrew-1234: Ah ok. Yeah my intent was the case where if no rows are returned, the export continues on until building the package metadata and eventually errors e.g. because project is nil. I thought it would be ok to check with .empty? because it does a minimal check

SELECT 1 AS one FROM \"audio_events_tags\" LIMIT $1"

but I realise now I  could also just check after the find_each is run if rows processed was 0

atruskie: >but I realise now I could also just check after the find_each is run if rows processed was 0

Yeah, do that

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/table/observation.rb
andrew-1234: do you mean make a User::full_name method that just returns user_name, since we dont have the information ?

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/descriptor/resource.rb
andrew-1234: fixed i think by moving the schema loading to package_metadata , so resource is non operational 

and the constants are on the root camtrap dp module

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/table/observation.rb
atruskie: Yeah, kind of like machine_generated for provenance. 

It's about making sure logic doesn't get buried. If we're referencing a stub now we can track the dependency later.

Eventually I think we want to add user names to profiles some how.

lib/gems/baw-workers/lib/baw_workers/export/camtrap_dp/descriptor/resource.rb
atruskie: Yeah recent change inhttps://github.com/QutEcoacoustics/baw-server/commit/405a87d3b79cdbcae399734e3bd0815acd382d29 looks good

PR #1032, opened 2026-08-09
fix: redirect to client home_url after confirming password reset

andrew-1234:
# fix: redirect to client home_url after confirming password reset

Prevent users getting stuck in `api` subdomain after resetting their password. 

## Changes

Add passwords controller and tests.

## Problems

Silent breaking change, tracked in https://github.com/QutEcoacoustics/baw-server/issues/1036

## Final Checklist

### DIFF
--- a/app/controllers/users/passwords_controller.rb
+++ b/app/controllers/users/passwords_controller.rb
@@ -1,0 +1,37 @@
+# frozen_string_literal: true
+
+class Users::PasswordsController < Devise::PasswordsController
+  # https://github.com/heartcombo/devise/blob/v5.0.3/app/controllers/devise/passwords_controller.rb
+  #
+  # We want to redirect to `client_routes.home_url` after a password reset, but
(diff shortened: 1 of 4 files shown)


atruskie (approved):
OK; Great simple change.

However, we do introduce a silent breaking change. Because we redirect to the client, the flash messages will be silently discarded. The client currently does not support showing flash messages so we can't fix this immediately. Not seeing if the operation was successful or not could be _really_ hostile for users.

I'd say we merge this to resolve the immediate issue, but also log tracking issues:

- one on the server to track this breaking change
- one on the client to add support for showing flash messages from the server.

These both may not be implemented because of our upcoming keycloak integration, but for now we have to at least track the problem

andrew-1234:
Created issues:
- https://github.com/QutEcoacoustics/baw-server/issues/1036
- https://github.com/QutEcoacoustics/workbench-client/issues/2679

PR #1046, opened 2026-09-07
feat(reports): add reports/tag_rate endpoint

### LINKED ISSUE #973: Tag rate report
Our reporting feature (#653) is almost ready. We've identified a need for a new report: one that reports average tag rates per unit time, filtered by successful analysis, per bucket.

The idea here is to report a normalised tag detection rate that accounts for gaps in audio or analysis coverage. The idea is tag detection rates should be comparable across different effort regimes.

we we're reporting:

    - unique tag detections per aligned minute of audio
         - controls for multiple analyses
         - controls for duplicate analyses
         - does not control for multiple analyses done over varying windows
         - varying analysis window lengths (5 seconds/3 seconds)

questions:

- what size is the aggregate window?
  - 60 seconds always
- Is it constrained to be less than bucket_size?
     - yes, but that's also non-issue with a fixed rate, probably still need to validate bucket_size < 60
- do we collapse any successful analysis into a success per recording? or not? or report it separately?
     - yes
- do we collapse duplicate tags?
     - yes
- do we just add this to the tag_frequency api?     
    -  unsure - depends on performance

Here is an existing tag frequency structure we'd adapt:

```bash
curl --json '{"filter":{"sites.id":{"eq":1234}},"options":{"bucket_size":"day"}}' -H 'Authorization: Token token="xxxxxx"' https://api.ecosounds.org/reports/tag_frequency
```

```jsonc
{
  "meta": {
    "status": 200,
    "message": "OK",
    "filter": {
      "sites.id": {
        "eq": 3964
      }
    }
  },
  "data": [
    {
      "bucket": [
        "2023-05-08T00:00:00.000Z",
        "2023-05-09T00:00:00.000Z"
      ],
      "tags": [
        {
          "events": 1,
          "tag_id": 3220
        }
      ]
    },
    {
      "bucket": [
        "2023-05-09T00:00:00.000Z",
        "2023-05-10T00:00:00.000Z"
      ],
      "tags": []
    },
    {
      "bucket": [
        "2023-05-10T00:00:00.000Z",
        "2023-05-11T00:00:00.000Z"
      ],
      "tags": []
    },
    {
      "bucket": [
        "2023-05-11T00:00:00.000Z",
        "2023-05-12T00:00:00.000Z"
      ],
      "tags": []
    },
    // snip
    {
      "bucket": [
        "2024-10-26T00:00:00.000Z",
        "2024-10-27T00:00:00.000Z"
      ],
      "tags": [
        {
          "events": 1,
          "tag_id": 3220
        },
        {
          "events": 7,
          "tag_id": 3221
        },
        {
          "events": 1210,
          "tag_id": 3217
        },
        {
          "events": 105,
          "tag_id": 3216
        },
        {
          "events": 169,
          "tag_id": 3218
        }
      ]
    }
  ]
}

### PR DESCRIPTION
**Pull request by andrew-1234**
# feat: add reports/tag_rate endpoint

Add a new `POST /reports/tag_rate` endpoint that returns normalised tag
detection rates grouped by site and configurable time bucket. Each group
includes recorded and analysed audio minutes, successful analysis job IDs,
manually reviewed minutes, and per-tag detected minutes split by manual and
analysis tagging sources.

Closes #973.

# Changes

- Add the `tag_rate` route and action to `ReportsController`
- Add `Api::Reporting::TagRate` with CTEs for calculating recording
  effort, analysed audio, manual tagging effort, analysis jobs, and per-tag
  detection rates
- Add `AudioRecording.recording_range_arel` for range-based report queries
- Add shared Arel support for JSONB aggregation and related PostgreSQL nodes
- Generalise the `Api::Reporting::Bucketer` and update coverage, tag accumulation, and
  tag frequency reports to use the shared behaviour
- Add API, request, routing and permission specs for `tag_rate`

## Problems

No problems identified.

## Visual Changes

No visual changes.

## Final Checklist

### CODE REVIEW

File: app/controllers/reports_controller.rb [hunk 1]
```diff
@@ -6,6 +6,28 @@ class ReportsController < ApplicationController
   include Api::Reporting
   include ResultFormatters
 
+  # POST /reports/tag_rate
+  # Returns a structured report of normalised tag rates
+  # Accepts a filter objectd where:
```
atruskie:
```suggestion
  # Accepts a filter object where:
```

app/controllers/reports_controller.rb
```diff
@@ -6,6 +6,28 @@ class ReportsController < ApplicationController
   include Api::Reporting
   include ResultFormatters
 
+  # POST /reports/tag_rate
+  # Returns a structured report of normalised tag rates
```
atruskie:
```suggestion
  # Returns a structured report of tag counts and
  # audio or analysis coverage for the same bucket.
  # The pattern is to report various numerators and demoninators
  # so callers can estimate different rates of detections based
  # on the amount of "work" also present in the bucket.
  # This endpoint is also unique because all counts are reported
  # per minute and de-duplicate tag detections in each minute.
```

app/controllers/reports_controller.rb
```diff
@@ -6,6 +6,28 @@ class ReportsController < ApplicationController
...
+  def tag_rate
+    do_authorize_class(:filter, AudioRecording)
+
+    base_query = Access::ByPermissionTable.audio_recordings(current_user, level: Access::Permission::READER)
+
+    tag_rate_template = TagRate.new(report_options)
+
+    results, opts = execute_report(
+      base_query:,
+      model: AudioRecording,
```
atruskie, line 39: Major possible issue: this report should be targeting the `AudioEvent` model. Discuss?

Filtering downwards through the hierarchy remains poorly defined, and I imagine most of our most interesting filtering will be done on audio events.

atruskie: Reinforcing this: this endpoint is meant to be a more complex/expensive version of the tag_frequency API which targets `AudioEvent` as well. That alone is enough of a smell.

spec/requests/reports/tag_rate_spec.rb
```diff
@@ -0,0 +1,345 @@
...
+        before do
+          create(:audio_event_using_tag, audio_recording:, creator:, tag: other_tag,
+            start_time_seconds: 1900, end_time_seconds: 1905)
+        end
+
+        it 'returns a tag result for all tags on the recording, not just the filtered tag' do
+          post '/reports/tag_rate', params: { options: { bucket_size: :day }, filter: {} }, **api_headers(writer_token)
+          expect_success
+
+          bucket = expected_data.first
```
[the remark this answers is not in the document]
andrew-1234: fixed

app/modules/api/reporting/tag_rate.rb
```diff
@@ -0,0 +1,315 @@
...
+          .except(:select, :order, :limit, :offset)
+          .reselect(
+            AudioRecording.recording_range_arel.as(RECORDING_RANGE),
+            AudioRecording.arel_table[:id].as('audio_recording_id'),
+            AudioRecording.arel_table[:site_id]
+          )
+          .arel
+      end
+
+      # Distinct successful analysis job ids per recording.
```
atruskie:
```suggestion
      # Distinct successful analysis job item ids per recording.
```

app/modules/api/reporting/tag_rate.rb
```diff
@@ -0,0 +1,315 @@
...
+      def analysed_recordings_cte
+        aji = AnalysisJobsItem.arel_table
+        job_ids = aji[:analysis_job_id].array_agg
+        job_ids.distinct = true
+
+        aji
+          .project(RECORDINGS[:audio_recording_id], job_ids.as('successful_analysis_job_ids'))
```
atruskie: missing `item` in all these definitions, but the most import one is the alias name - since that can leak

app/controllers/reports_controller.rb
```diff
@@ -6,6 +6,28 @@ class ReportsController < ApplicationController
...
+      template: tag_rate_template,
+      projections: { tags: tag_rate_template.tags_summary }
```
atruskie: is there a reason we included one of the default projections (`tags`) and omitted all the other ones?

I'm in favour of having them all here as they self-document the shape of each bucket.

app/modules/api/reporting/tag_rate.rb
```diff
@@ -0,0 +1,315 @@
...
+
+      # Return the total minutes of recorded and analysed audio per site/bucket.
+      def total_minutes_cte
+        RECORDING_RANGE_SLICES
+          .project(
+            RECORDING_RANGE_SLICES[:bucket],
+            RECORDING_RANGE_SLICES[:site_id],
+            total_minutes.as('total_minutes'),
+            total_analysed_minutes.as('total_analysed_minutes')
+          )
```
atruskie, line 155: nit: successful analyzed minutes, not just analyzed minutes.

I don't think it matters, but we do have both concepts, so confusing them is, well confusing.

app/modules/api/reporting/tag_rate.rb
```diff
@@ -0,0 +1,315 @@
...
+          .group(RECORDING_RANGE_SLICES[:bucket], RECORDING_RANGE_SLICES[:site_id])
+      end
+
+      # @return [Arel::Nodes::Division] total minutes of recorded audio (ceiled)
```
atruskie:
```suggestion
      # @return [Arel::Nodes::Division] total minutes of recorded audio (rounded up)
```

app/modules/api/reporting/tag_rate.rb
```diff
@@ -0,0 +1,315 @@
...
+        # parameterized nested-loop plan that uses the audio_recording_id index. A
+        # direct inner join is logically equivalent but was slower by ~10
+        # seconds in the report benchmark.
+        events_sub = events
+          .project(events[:id], events[:start_time_seconds], events[:audio_event_import_file_id])
+          .where(events[:audio_recording_id].eq(RECORDINGS[:audio_recording_id])).skip(0)
+
+        events_sub_table = Arel::Nodes::TableAlias.new(events_sub, 'audio_events')
+
+        event_start_at = RECORDINGS[RECORDING_RANGE].lower + events_sub_table[:start_time_seconds].seconds
```
atruskie, line 217: ok I think you've already addressed this, but why aren't we using https://github.com/QutEcoacoustics/baw-server/blob/34e23090b33e553c75d094b5f58f3d818b72c159/app/models/audio_event.rb#L129-L134

?

app/modules/api/reporting/tag_rate.rb
```diff
@@ -0,0 +1,315 @@
...
+            tagging_source.as('tagging_source')
+          )
+          .join(Arel::Nodes::Lateral.new(events_sub_table)).on(Arel.sql('true'))
+          .join(taggings).on(taggings[:audio_event_id].eq(events_sub_table[:id]))
+          .join(imports, Arel::Nodes::OuterJoin).on(imports[:id].eq(events_sub_table[:audio_event_import_file_id]))
+          .distinct
+      end
+
+      # Unique minutes with any manual event per bucket. We use this to provide
+      # a crude estimate of the manual tagging effort, since we don't have any
```
atruskie:
```suggestion
      # a crude estimate of the manual analysis effort (e.g. manual tagging), since we don't have any
```

spec/api/reports/tag_rate_spec.rb
```diff
@@ -0,0 +1,185 @@
...
+describe 'reports', type: :request do
+  create_entire_hierarchy
+
+  sends_json_and_expects_json
+  with_authorization
+
+  let(:skip_automatic_description) { true }
+
+  def self.response_body_schema
+    Api::Schema.standard_array_response(
```
atruskie, line 14: i know this is our convention for reports,  but having api definition in test files, where for the rest of our models it's in with our models is very inconsistent. TODO, clean this up later

spec/api/reports/tag_rate_spec.rb
```diff
@@ -0,0 +1,185 @@
...
+    post 'Gets tag rate report per site and time bucket' do
+      tags 'reports'
+      consumes 'application/json'
+      produces 'application/json'
+
+      description <<~DESCRIPTION
+        # Returns tag detection counts per site and time bucket.
+        # The `options` parameter specifies the bucket size (day, week, month, or year).
+        # The optional `filter` parameter is applied to audio recordings.
+        # Results only include audio recordings the user has reader access to.
```
atruskie: Two notes, why are these commented strings? And should probably include some comment on the effort stats reported (audio recording / analysis minutes)

spec/api/reports/tag_rate_spec.rb
```diff
@@ -0,0 +1,185 @@
...
+        end
+      end
+
+      response '200', 'filters audio recordings by tag' do
+        let(:request_body) do
+          {
+            options: { bucket_size: 'day' },
+            filter: { 'tags.id': { eq: tag.id } }
+          }
+        end
```
atruskie, line 108: Going right back to the comment in reports controller, I have no idea what this will produce. Any whole recording is reported that has  a tag anywhere in it?

spec/permissions/reports_spec.rb
```diff
@@ -26,6 +28,28 @@
...
+            range: [
+              audio_recording.recorded_date.utc.at_beginning_of_day.as_json,
+              (audio_recording.recorded_date.utc.at_beginning_of_day + 1.day).as_json
+            ],
+            tags: [{ tag_id: tag.id, detected_manual_minutes: 0, detected_analysis_minutes: 1,
+                     detected_combined_minutes: 1 }],
+            analysis_ids: [second_analysis_job.id],
+            total_minutes: audio_recording.duration_seconds / 60.0,
+            manual_events_minutes: 0,
+            total_analysed_minutes: audio_recording.duration_seconds / 60.0 }
```
atruskie: OK, unless you're actually testing the variance of these stats based on the user, you don't have to, and should not, test the definitions of the porjections here. Maybe keep one or two just to test the data is present, but even those should be a `is_an_number` assertion rather than an exact definition

spec/requests/reports/tag_rate_spec.rb
atruskie, line 1: ok honestly, i found it difficult to fully load this test into my head. 

the creation code was too verbose - i probably needed a diagram

and the modification of the expected response was too compact - i couldn't keep that in my head. 

this is a neutral comment though: if I were working with this code it'd probably be all right.

atruskie (changes requested):
OK, a potentially annoying change: `detected_manual_minutes` should probably be named: `detected_imports_minutes` or `detected_in_imports_minutes`.   Maybe even `detected_unknown_minutes`? 

Our imports can be manual annotations, or analyses that are done offline. If we don't have analysis information we don't know how they were generated - we can't infer they're manual.

All in all great work. The major discussion point is that the query is not based off AudioEvent. Other than all the (very complex) calculations appear to make sense!

app/modules/api/reporting/tag_rate.rb
```diff
@@ -0,0 +1,315 @@
...
+          .project(
+            RECORDINGS[:site_id],
+            taggings[:tag_id],
+            tagged_minute.as('tagged_minute'),
+            tagging_source.as('tagging_source')
+          )
+          .join(Arel::Nodes::Lateral.new(events_sub_table)).on(Arel.sql('true'))
+          .join(taggings).on(taggings[:audio_event_id].eq(events_sub_table[:id]))
+          .join(imports, Arel::Nodes::OuterJoin).on(imports[:id].eq(events_sub_table[:audio_event_import_file_id]))
+          .distinct
```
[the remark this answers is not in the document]
atruskie: It's probably worth making the change. 

I'd say we shouldn't import events for non-successful items (and we don't: https://github.com/QutEcoacoustics/baw-server/blob/34e23090b33e553c75d094b5f58f3d818b72c159/app/models/analysis_jobs_item/state_machine.rb#L284-L290) 

But this rule is not enforced anywhere outside of this line, so the database could still have results in it (say from a manually run import job) that violate this assumption

app/controllers/reports_controller.rb
andrew-1234: fixed

app/modules/api/reporting/tag_rate.rb
atruskie: I misread the intent because there is no such thing as an successful analysis job (`successful_analysis_job_ids`). My comment is mostly wrong, but maybe change that alias

spec/api/reports/tag_rate_spec.rb
andrew-1234: fixed

app/modules/api/reporting/tag_rate.rb
andrew-1234: since audio_recording table isn't in the from clause

